PR #109, opened 2014-06-17
Extending newman to have alternative response handler

jaetask:
I have extended the newman binary to receive the filename of a response handler. The response handler must be in the same folder as the supplied ones. this provides consumers the ability to override how tests are reported.


czardoz:
Closing this PR, since it's too old and does not fit into Newman's current architecture

PR #187, opened 2015-02-23
Convert access to responseHeaders with correct camelCase

ViBiOh:
Convert access to responseHeaders with correct camelCase for custom header name


abhijitkane:
You can use `postman.getResponseHeader(headerName)` to get response headers in a case-insensitive format.

PR #225, opened 2015-07-03
Feature -  Adding support for polling an API with repeated requests.

digital-abyss:
An example use case for this feature is when you need to poll an API for status.  

:e,g,

POST /some/uri/  ---> this returns some 'request received
successfully' response, and kicks off some long running process on the
server.

GET /some/uri ---> this returns some status: eg 'in progress' |
'completed' | 'failed'

You can now add the following fields to your json test in your
collection:

iterationsUntilFail ---> defines the maximum number of times you want to retry the call
requestDelay ---> the delay to wait between repeated requests in
ms.

Newman will repeatedly hit the URI until all response tests pass, or the number
of iterations is reached.  By setting up your tests properly, (eg, check
for 'completed'), you can now essentially wait until a desired response
is received.
Note that interim failures will not be counted towards the test counts.
Only the last iteration's tests will be added in.

------------------------
I wasn't sure which branch to make the PR against, so chose master, but can retarget if necessary.


czardoz:
@abhijitkane what do you think about this feature? I am -1 on this because Newman can be wrapped with an external script with the ``-i`` and ``-s`` options to achieve the same result. It also introduces more use of the Globals, which will make getting rid of them that much harder.

czardoz:
@digital-abyss Have you tried wrapping Newman in an external script? Is there any particular reason this functionality has to exist inside Newman?

abhijitkane:
@czardoz Agree. Given the fact that feature-parity with the collection runner is important, we'll hold off till we can implement this in Postman as well.

digital-abyss:
@czardoz -- can you elaborate on how to achieve the same effect with -i and -s options ?  

I think it is a natural next step for API test writers to want to be able to have looping and 'if' style constructs around a single API.  This feature has been a major driver of adoption of postman + newman + jetpacks for my team.

For my organization, the real power comes in the ability to automate tests.  I have 83 collections, most with over 30 API calls.  Most of these collections rely on asynchronous behaviour.  It makes very little sense to try and script around newman and increase the number of collections as well as force dependencies between collections.  

Newman also has a non-zero startup cost.  Besides creating maintenance issues, having to split logically grouped collections will now incur a performance penalty for any repeated API.

If there's an issue with the implementation of the feature (in terms of exposing more globals), I'm happy to try and rework it if there are some suggestions to improve it.  

I understand the importance of feature parity, but don't think I can contribute to the collection runner ?

MustafaHaddara:
+1 on this feature. Being able to test asynchronous API calls is indispensable to our testing framework.

zAlbee:
This feature is best within Newman (and Postman/the collection runner).

We use collections to group and sequence REST calls/tests. It doesn't make sense to repeat the collection externally, because then you are asking us to break up a single logical collection into multiple collections and then somehow group and sequence our collections (which Postman has no ability to do). Instead of adding another level of indirection, I think it's most natural to mark a single request or test as repeating, while keeping its place in the same collection, which is what this PR does.

czardoz:
We can now use `postman.setNextRequest()` to poll a single endpoint. Closing this, because Newman is now v3, and this PR cannot be merged. Thanks a lot, @digital-abyss :cake:

PR #241, opened 2015-08-04
Implemented AWS Signature V4 authentication

czardoz:


abhijitkane:
The x-amz-date can also be replaced by the date header - 
http://docs.aws.amazon.com/general/latest/gr/sigv4-date-handling.html

Can you add support for that?

czardoz:
Fixed the handling of the date header, and also picking up the service name from the collection

PR #256, opened 2015-09-09
Make newman Node 4.0.0 compatible

kevincennis:
A few changes:

### 	Update jsdom to 6.3.0 

This version no longer depends on contextify, which was using an older version of the nan module that wasn't 4.0.0 compatible.

### Don't access `arguments` or `caller` properties on `Function.prototype`

ES6 has deprecated these properties, and replaced them with accessors that just throw a `TypeError` if you try to touch them.

My workaround is to simply copy *entire file descriptors* instead of using direct assignment. This way, you maintain enumerability, setters/getters, etc.

### Fix EPIPE error during version check

Seems like `iterationRunnerOver` was getting called before the `exec` callback, so I just added some flags to make sure we wait for that to finish before invoking the `Newman.execute` callback.

------

Please note that if you decide to merge this, it will definitely need to be a *major* version bump because jsdom >=4.x.x does not work with older versions of Node. So I think you'd need to draw a line in the sand that newman 2.x.x is for Node 4.0.0 or higher.


czardoz:
This is great, thanks @kevincennis! :cake:  
We already have a branch for Node v4.0.0: https://github.com/postmanlabs/newman/tree/feature/node-v4-support 
Could you open up this PR against that branch?

kevincennis:
Yup. Submitted #257 

Guess I probably should have looked around a bit before opening this...

¯\_(ツ)_/¯

PR #257, opened 2015-09-09
More Node v4 stuff

kevincennis:
Two changes:

### Don't access `arguments` or `caller` properties on `Function.prototype`

ES6 has deprecated these properties, and replaced them with accessors that just throw a `TypeError` if you try to touch them.

My workaround is to simply copy entire file descriptors instead of using direct assignment. This way, you maintain enumerability, setters/getters, etc.

### Fix EPIPE error during version check

Seems like `iterationRunnerOver` was getting called before the `exec` callback, so I just added some flags to make sure we wait for that to finish before invoking the `Newman.execute` callback.

### DIFF
--- a/src/Newman.js
+++ b/src/Newman.js
@@ -25,6 +25,8 @@ var Newman = jsface.Class([Options, EventEmitter], {
      * @param {object} Newman options
      */
     execute: function(requestJSON, options, callback) {
+        var checking = false,
+            onChecked = null;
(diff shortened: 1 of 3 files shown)


kevincennis:
This is failing because the new `jsdom` breaks on older Node versions. Happy to update `.travis.yml` if you want me to.

czardoz:
Nah, I'll take care of Travis. Thanks again! :)

PR #305, opened 2015-12-18
Feature/config file

ajsager:


bin/newman
```diff
@@ -21,10 +21,25 @@ var color   = require('cli-color'),
...
 
+function parseConfigFile() {
+	var config,
+		option;
+
+	Helpers.validateConfigFile(program.configFile);
+	config = Helpers.validateAndReturnConfig(program.configFile);
+
+	for (option in config) {
+		if (!program[option]) {
```
shamasis: `!program[option]` would pass for `false` and zero. Not sure whether commander parses to Boolean and number, but surely checking for own property or inequality with null and undefined is better.

bin/newman
```diff
@@ -55,17 +70,22 @@ function parseArguments() {
 		console.log('  A collection file or URL is mandatory');
 		console.log('  Examples:');
 		console.log('');
+		console.log('    newman -C CONFIG_FILE');
 		console.log('    newman -c POSTMAN_COLLECTION');
 		console.log('    newman -u POSTMAN_COLLECTION -e POSTMAN_ENVIRONMENT');
 		console.log('    newman -c POSTMAN_COLLECTION -f SAMPLE_FOLDER');
 		console.log('');
-		console.log('  For more documentation, issues and examples head over to https://github.com/a85/Newman');
+		console.log('  For more documentation, issues and examples head over to https://github.com/postmanlabs/newman');
```
shamasis: Ah. @pjsager nice catch. :+1:

src/utilities/Helpers.js
```diff
@@ -12,6 +12,55 @@ var jsface = require('jsface'),
...
+        try {
+            config = require(file);
+        }
+        catch (RefrenceError) {
+            Errors.terminateWithError("The config file specified is not a valid JSON file");
+        }
+
+        Object.keys(config).forEach(function (option) {
+            if (!validOptions[option] || (typeof config[option] !== validOptions[option])) {
+                Errors.terminateWithError("Config JSON has invalid options");
```
shamasis: Do we want to make the config file this strict? What if we allow these, but merge only the relevant ones?

src/utilities/Helpers.js
```diff
@@ -12,6 +12,55 @@ var jsface = require('jsface'),
...
+                insecure: "boolean",
+                tls: "boolean",
+                encoding: "string",
+                exitCode: "boolean",
+                outputFile: "string",
+                testReportFile: "string",
+                import: "string",
+                pretty: "boolean",
+                html: "string"
+            },
```
shamasis: Now that's one more place to maintain the list of config. Not that it's too much pain, but is there a way to pick this up straight from commander definition? :-p

shamasis:
@ajsager - nicely done. :-)
there are some merge conflicts though.

src/utilities/Helpers.js
czardoz: I prefer strict.. There's never unexpected behavior that way, and it forces people to look at exactly what options are supported

shamasis:
I also see that no new tests are added for the config file feature.

ajsager:
These changes were from @elssar.  I mean to make a future change, but have not made any commits yet.

shamasis:
@ajsager - I guess that's why he didn't send a PR. We will wait for @elssar to send a fresh PR then.

PR #334, opened 2016-02-23
Add port and protocol for 'aws4' request inputs.

harshavardhana:
A patch related to this is pending upstream
   - https://github.com/mhart/aws4/pull/21

With this change it would enable `newman` tool to be used
with AWS S3 compatible servers implementing AWS Signature
Version '4'.


czardoz:
:+1: 

I'll merge this when a new version of 'aws4' is released.

PR #336, opened 2016-02-24
awsSignv4: Collection properties should be 'service' not 'serviceName'

harshavardhana:
Collection properties for helper attributes lists 'service' not 'serviceName'.
```
"helperAttributes": {
	"id": "awsSigV4",
...
...
	"service": "s3",
	"auto": true,
	"saveHelper": true
},
```

This fixes a bug where 'execute-api' is used as part of Signature calculation
and failing.


czardoz:
:cake:

PR #342, opened 2016-03-04
removed check for empty strings from CSV to be consistent with Postman

bwolski:
Hi, my requests sometimes require empty query parameter values, which I am importing from CSV. Postman does not have an issue with importing empty strings from CSV, but Newman was setting these values as Null. 

I'm sure there's a better way to solve this issue, but this little change works for me, so I figure I'd share.

Thanks!
Benjamin


czardoz:
Thanks!

Issue #362, opened 2016-03-21
Redirects are automatically avoided 


beatsbears:
In version 2.0.3 all redirects are currently being avoided, making the -R flag redundant.  This should be changed so that redirects are automatically followed, unless the -R flag is supplied.

czardoz:
@beatsbears Thanks a lot for the PR!

I'm not able to reproduce this issue however. I have this collection: https://www.getpostman.com/collections/0b500894597958ffce2f which has a request that does a single redirect.

Without the "-R" option, I get the 200 response (which is expected):
```
$ newman -u https://www.getpostman.com/collections/0b500894597958ffce2f

Iteration 1 of 1
The return code I got is:  200
200 4173ms r1 [GET] http://httpbin.org/redirect/1
    ✗ Returned 302

Summary:

Parent                   	Pass Count	 FailCount
-------------------------------------------------------------
Collection redirectTest  	         0	         1

Total                    	         0	         1
```

When I run it with the "-R" option, it avoids the redirects, and shows the actual 302 response code:
```
$ newman -u https://www.getpostman.com/collections/0b500894597958ffce2f -R

Iteration 1 of 1
The return code I got is:  302
302 581ms r1 [GET] http://httpbin.org/redirect/1
    ✔ Returned 302

Summary:

Parent                   	Pass Count	 FailCount
-------------------------------------------------------------
Collection redirectTest  	         1	         0

Total                    	         1	         0
```

beatsbears:
Interesting I'll look into it a bit more if I get some time.  The only major difference that I'm seeing right now is that in my call I'm doing a POST, which gets redirected to a GET request rather than a GET redirecting to another GET. Looking at the HTTP spec, it doesn't sound like this is the 'correct' way to do things, but it is how it's commonly handled. 
[More here.](http://stackoverflow.com/questions/8138137/http-post-request-receives-a-302-should-the-redirect-request-be-a-get )

Besides that, we're using different servers and I have a few extra request headers such as Origin and Referer

shamasis:
@czardoz can we come to a conclusion on this?

kunagpal:
@shamasis @czardoz Are we good to close this issue?

shamasis:
I think our goal is to closely mimic XHR behaviour so that requests are as close to the browser as possible. closing this.

PR #363, opened 2016-03-22
fixed -R redirect flag, redirects were disabled

beatsbears:
Addressing issue #362.  Follow redirects was hardcoded to false regardless of the -R flag.

### DIFF
--- a/src/runners/RequestRunner.js
+++ b/src/runners/RequestRunner.js
@@ -229,8 +229,7 @@ var RequestRunner = jsface.Class([Queue, EventEmitter], {
         RequestOptions.url = request.transformed.url;
         RequestOptions.method = request.method;
         RequestOptions.headers = Helpers.generateHeaderObj(request.transformed.headers);
-        RequestOptions.followAllRedirects = false;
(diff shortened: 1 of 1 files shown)


beatsbears:
I'm not sure if this test is correct, I think the correct response code should be 200 unless the -R flag is passed.

czardoz:
@beatsbears That's almost the same test that I shared with you in #362.

The test expects that no redirects happen when "-R" flag is set to true. Since the response code from HTTPBin is "200" it means that Newman followed the redirect (which is causing the test to fail).

beatsbears:
Ah yes, you're right.  My personal tests, with a POST -> 302 GET do pass with my changes, but the standard GET -> 302 GET case is now failing whether or not the -R flag is given.

czardoz:
Looks like the build's fixed :cake: 
Is this ready for merge?

beatsbears:
Hope so!

czardoz:
@shamasis this makes the redirect behavior as compatible with browsers as possible.

The only difference now is that Browsers will do:
```
POST + 302 => GET request 
```
But Newman will
```
POST + 302 => POST
```

PR #368, opened 2016-03-29
Add an option (-S, --noTestSymbols) to disable symbols in test output…

### LINKED ISSUE #344: Feature-Request: Parameter to change result symbol to text
It would be great if you add a command line parameter to replace the symbols for Passed (✔) or Failed (✗) with the text like "PASS" or "FAIL".
They are replaced with confusing symbols (�) in different charsets.

chrisscott:
… and use PASS and FAIL instead. Fixes #344.


czardoz:
:cake:

PR #388, opened 2016-04-15
Add help text for current default request timeout

ramblinwreck35:
Just an idea, as a recent experience in our test environment bamboozled us by not realizing the difference in the request timeout between postman and newman (programatically set in CollectionRunner execute() towards the bottom). This value hasn't changed since April 13th of last year (0d6cf54dae831fd9ce3ce829ee282d0f7ecfd1ac), maybe documentation matching it may be helpful to some?


czardoz:
Thanks!

PR #420, opened 2016-06-07
Improve JUnit report for jenkins

xjodoin:
Jenkins use the classname attribute to display tests informations

http://nelsonwells.net/2012/09/how-jenkins-ci-parses-and-displays-junit-output/


czardoz:
These concerns have been addressed in Newman v3 :tada: 
Do try it out!

PR #422, opened 2016-06-08
Fix intial assignment of Global.globalJson

darcymiranda:
The global environment variables file isn't getting correctly assigned to Globals.globalJson. 

It seems that the way that it successfully worked originally (referencing the values property) was commented out for some reason.


darcymiranda:
Global environment variables test data tests/g3.json seems to be old? Exporting globals from Postman has the file in the same format as the environment variables. I'll fix the test

PR #423, opened 2016-06-08
Fixed Global Environment Variable Usage to Use the New Format

darcymiranda:
The g3.json intergration test is 2 years old so I assume the format is wrong considering the errors that I am getting using the exported global environment variables on the latest version of postman.

I kept support for the old format and added a new test for the new format.


czardoz:
@darcymiranda Thanks a lot for this!

I don't think the format has changed though. What errors are you seeing?

darcymiranda:
@czardoz This is our mistake. The "Global Environment" variables that was supplied to me by my colleague actually ended up being a save of the normal Environment Variables.

I will close.

PR #434, opened 2016-07-03
Add proxy support

jerryzhaoau:


czardoz:
Thanks! I don't actually want to add another option for setting proxies though. Currently, they can be set using the standard environment variables, `HTTP_PROXY=<url>` etc. 

That ensures that Newman picks up the proxy from the system settings. Is there any other use case for which you need this to be a command line option?

jerryzhaoau:
It would be more flexible when integrating with other systems in build pipelines. For example, it can be integrated to zaproxy to do a security scan for APIs.
Sometimes it may not possible to change system proxy in build server.

shamasis:
@jerryzhaoau this should go in newman v3.1+ (not coded in newman but in postman-runtime so that even native apps can have a way to configure this)

czardoz:
Closing this because it will be implemented in v3.1 :smile:

PR #445, opened 2016-07-18
Feature/v3 configuration

czardoz:
Better inheriting of configuration


lib/run.js
```diff
@@ -48,64 +66,69 @@ module.exports = function (options, callback) {
     options = _.isObject(options) ? _.clone(options) : {}; // we initialise this as object to avoid needless validations
```
shamasis: this check is no longer needed

PR #450, opened 2016-07-20
emit as single object argument

shamasis:
While converting runtime callbacks to events, transform the arguments into a single object.

### DIFF
--- a/lib/summary.js
+++ b/lib/summary.js
@@ -1,5 +1,4 @@
 var _ = require('lodash'),
-    CollectionItem = require('postman-collection').Item,
     SerialiseError = require('serialised-error'),
     RunSummary;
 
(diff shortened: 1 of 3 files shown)


lib/summary.js
```diff
@@ -86,19 +90,21 @@ RunSummary.trackEvent = function (events, emitter) {
...
+                error: err,
+                cursor: o.cursor,
+                parent: o.item && item.__parent && item.__parent.__parent
             });
         });
 
-        emitter.on(event, function (err, cur) {
-            var item = RunSummary.sniffItemFromArgs(arguments),
+        emitter.on(event, function (err, o) {
+            var item = o.item,
```
czardoz: What's `o`?

PR #452, opened 2016-07-20
Support for V1 Collections

czardoz:
Also:

- Added philosophy to BETA.md
- Support for JSON data files


lib/run.js
```diff
@@ -73,7 +101,7 @@ module.exports = function (options, callback) {
     // validate all options. it is to be notet that `options` parameter is option and is polymorphic
     (!callback && _.isFunction(options)) && (
         (callback = options),
-        (options = {})
+            (options = {})
```
shamasis: This indentation is needless.

PR #455, opened 2016-07-21
Enabled [-d | --iteration-data] recognition for Newman v3

kunagpal:
@czardoz @shamasis

### DIFF
--- a/lib/run.js
+++ b/lib/run.js
@@ -85,20 +85,22 @@ var _ = require('lodash'),
         },
         environment: externalLoader.bind(this, 'environment'),
         globals: externalLoader.bind(this, 'globals'),
-        data: externalLoader // todo: use a csv parser to allow csv files.
+        iterationData: externalLoader.bind(this, 'iterationData') // todo: use a csv parser to allow csv files.
     };
 
 /**
  * @param {Object} options
  * @param {Collection|Object} options.collection
  * @param {Object} options.environment
+ * @param {Object} options.globals
+ * @param {Object} options.iterationData
  * @param {Array|String} options.reporters
  * @param {Function} callback
  *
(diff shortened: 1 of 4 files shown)


lib/run.js
```diff
@@ -85,20 +85,22 @@ var _ = require('lodash'),
         },
         environment: externalLoader.bind(this, 'environment'),
         globals: externalLoader.bind(this, 'globals'),
-        data: externalLoader // todo: use a csv parser to allow csv files.
+        iterationData: externalLoader.bind(this, 'iterationData') // todo: use a csv parser to allow csv files.
```
czardoz: We cannot use `externalLoader` to load iteration data, because it does some unnecessary transformations (for this case). 
Please use `util.fetch` directly.

czardoz:
@kunagpal also test your change out, by actually running a collection that uses iteration data.
For example: I tested this out by creating a data file: 

```
[
 {
  "var": "one"
 },
 {
  "var": "two"
 }
]
```
and a collection: https://www.getpostman.com/collections/c18cca0018057e550055

lib/run.js
```diff
@@ -135,7 +137,7 @@ module.exports = function (options, callback) {
             environment: options.environment,
             globals: options.globals,
             entrypoint: options.folder,
-            data: options.data, // todo: should this have a better name?
+            data: options.iterationData, // todo: should this have a better name?
```
czardoz: We no longer need the todo there :)

shamasis:
Where's the iterationCount?

czardoz:
`iterationCount` was added some time ago: https://github.com/postmanlabs/newman/pull/455/files#diff-de78b50723ff596d7bd35665f36a6cbfR136

PR #460, opened 2016-07-25
Added version number to NSP test

shamasis:
This PR sends newman version to NSP for their DB usage


shamasis:
This is a trivial system test change + travis build has passed on Node v4 and for some weird reason, build is not starting on other node versions. Going ahead and merging it. + @czardoz

PR #462, opened 2016-07-25
v3 => separate cli reporter failure parser

shamasis:
This PR separates the function in CLI reporter that parses run failures and creates a CLI table. As such this can now be internally re-used to optionally format errors and also to make code look readable when CLI receives options


shamasis:
@czardoz - travis seems to be stuck. please test locally, review changes and merge if okay.

PR #465, opened 2016-07-25
v3 => check tty and turn off spinner

shamasis:
* check presence of CI or lack of TTY and turn off interactive spinner
* turn off column width checking when window size is not found

> *merge https://github.com/postmanlabs/newman/pull/462 first*

Easy way to test this PR is piping newman to cat
`./bin/newman.js run examples/sample-collection.json | cat`


shamasis:
Wrong target. Closing

PR #467, opened 2016-07-26
This change was not merged into v3

czardoz:
Which is why ignoreRedirects was not working :(


czardoz:
Merged as a part of #469

PR #471, opened 2016-07-26
Added collection `name` and `id` to run summary

kunagpal:
@czardoz @shamasis

### DIFF
--- a/lib/run.js
+++ b/lib/run.js
@@ -220,8 +220,10 @@ module.exports = function (options, callback) {
 
             // override the `done` event to fire the end callback
             callbacks.done = function (err) { // @todo - do some memory cleanup here?
+                var collectionDetails = _.pick(options.collection, ['id', 'name']);
+
                 emitter.emit('done', err, emitter.summary); // we now trigger actual done event which we had overridden
-                callback(err, emitter.summary);
+                callback(err, _.merge(emitter.summary, collectionDetails));
             };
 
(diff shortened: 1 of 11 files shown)


lib/run.js
```diff
@@ -220,8 +220,10 @@ module.exports = function (options, callback) {
 
             // override the `done` event to fire the end callback
             callbacks.done = function (err) { // @todo - do some memory cleanup here?
+                var collectionDetails = _.pick(options.collection, ['id', 'name']);
+
                 emitter.emit('done', err, emitter.summary); // we now trigger actual done event which we had overridden
-                callback(err, emitter.summary);
+                callback(err, _.merge(emitter.summary, collectionDetails));
```
czardoz: With this methodology, only the end consumer of the summary object will have access to these details.
Ideally, it would be better to pass these details as options to the `RunSummary` constructor, so that the summary always has them, and everyone can use them.

https://github.com/postmanlabs/newman/pull/471/files#diff-de78b50723ff596d7bd35665f36a6cbfR166

czardoz: Also, it might be good to add a check to ensure that these details are present in the summary, right around here:
https://github.com/postmanlabs/newman/pull/471/files#diff-de78b50723ff596d7bd35665f36a6cbfR166

shamasis:
@czardoz - made all the changes we had discussed. please review and merge

czardoz:
LGTM! Will try it out and merge in a bit.

PR #480, opened 2016-07-27
v3 => Response size output in CLI

shamasis:
- adds the response size using same formatting as Postman Canary beside every request
- adds total data received summary row of CLI

Out of context:
- move the timings in summary to `summary.timings object`
- compute average response time in summary
- adds the average response time to summary table row of CLI


lib/summary.js
```diff
@@ -42,11 +48,28 @@ RunSummary = function RunSummary (emitter, options) {
...
+    emitter.on('start', function () { summary.timings.started = Date.now(); });
     emitter.on('done', function (err) {
-        summary.completed = Date.now();
+        summary.timings.completed = Date.now();
         err && (summary.error = err);
     });
+
+    // accumulate statistics on requests
+    emitter.on('request', function (err, o) {
+        if (!(!err && o && o.response)) { return; }
```
czardoz:
```
if (err || !(o && o.response))
```
Would be better here

PR #481, opened 2016-07-27
Fixed two bugs

czardoz:
- Newman was not accepting Objects when given in the configuration
- Ensure that Newman sends a User-Agent header when fetching stuff

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -1,7 +1,8 @@
 var _ = require('lodash'),
     request = require('request'),
     parseJson = require('parse-json'),
-    fs = require('fs');
+    fs = require('fs'),
+    version = require('../package.json').version;
 
 module.exports = {
     /**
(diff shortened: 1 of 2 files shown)


lib/util.js
```diff
@@ -16,10 +17,10 @@ module.exports = {
         !callback && _.isFunction(options) && (callback = options, options = {});
         return (/^https?:\/\/.*/).test(location) ?
             // Load from URL
-            request.get({ url: location, json: true }, (err, response, body) => {
-                if (err) {
-                    return callback(err);
-                }
+            request.get({
+                url: location, json: true, headers: { 'User-Agent': 'Newman/' + version }
```
shamasis: why compute this `Newman/' + version every time??!! Just compute once on top. :-/

shamasis: How is this passing lint with multiple keys of object in one line???

lib/util.js
```diff
@@ -16,10 +17,10 @@ module.exports = {
...
+            }, (err, response, body) => {
+                if (err) { return callback(err); }
```
shamasis: What happened to the function end brace indentation here? Should it not be in new line?

czardoz: It is on a different line.

Issue #472, opened 2016-07-26
Support Client Side Certificate Authentication


anirupdutta:
So the API we are testing against requires us to use a certificate to authenticate. From what I have read it may not be something we will be able to incorporate with newman. It would be nice to have support for it.

shamasis:
Newman maintains feature parity with postman app. Need to check how this will work in postman app before implementing in Newman.

tdurden82:
In OSX Postman, the cert appears to be managed by Chrome (via Keychain).

It would be nice to tie/define a certificate in the environment so we can switch.

shamasis:
Now that we moved to NodeJS driven runtime, this feature has been bumped up in priority.
+ @numaanashraf @czardoz

The goal:
- understand the Chrome magic when it picks up the certificate
- possibly add better user experience around certificates
- implement it across the board

shamasis:
@tdurden82 - yes, either via environment or using an in-app certificate store (like our cookie store) should make the workflow better.

pawelka:
Hi, 

Here is my proposition to solve the problem:

```
diff --git a/lib/run/index.js b/lib/run/index.js
index 0234f94..ad328ce 100644
--- a/lib/run/index.js
+++ b/lib/run/index.js
@@ -94,7 +94,16 @@ module.exports = function (options, callback) {
             timeout: options.timeoutRequest ? { request: options.timeoutRequest } : undefined,
             requester: {
                 followRedirects: _.has(options, 'ignoreRedirects') ? !options.ignoreRedirects : undefined,
-                strictSSL: _.has(options, 'insecure') ? !options.insecure : undefined
+                strictSSL: _.has(options, 'insecure') ? !options.insecure : undefined,
+                certificateManager: _.has(options, 'certInfo') ? {
+                    getCertificateContents: function(domain, callback){
+                        if(_.has(options.certInfo, domain)){
+                            callback(null, options.certInfo[domain]);
+                        }else{
+                            callback('Certificate not found for domain' + domain, null);
+                        }
+                    }
+                } : undefined
             }
         }, function (err, run) {
             var callbacks = {},
```

Usage of above patch:
```
newman.run({
    ...
    certInfo: {
        "server1.com" : {
                                        pem: fs.readFileSync(certFile1), //reading content from file 
                                        key: fs.readFileSync(keyFile1),
                                        passphrase: 'password1'
                                   }
        },
        "server2.com" : {
                                        pem: fs.readFileSync(certFile2),
[...]

PR #482, opened 2016-07-27
Adds the option to pass in a client certificate and key.

tdurden82:
#472 
Adds the option to pass in a client certificate and key.


shamasis:
Now that we moved to NodeJS driven runtime, this feature has been bumped up in priority. The new implementation now has to go into Postman Collection SDK and the underlying postman-runtime.

Thank you @tdurden82 for the PR. I am afraid, that we will have to close it and pick this up as a feature implementation in our underlying postman-runtime module - thus bringing this to newman and all our apps.

PR #484, opened 2016-07-28
Added collection run summary object structural tests

kunagpal:
@czardoz @shamasis


shamasis:
We will use the alternate testing approach. A lot of functional tests need to go in instead of just a structural test.

PR #485, opened 2016-07-28
v3 => summary refactor

shamasis:


shamasis:
@czardoz merge #480 before this.

PR #498, opened 2016-08-02
v3 => beforeDone event and some code organisation

shamasis:
- adds `beforeDone` event for later consumption in reporters
- moves run.js config loading to a separate file

> **DO NOT DELETE BRANCH POST MERGE **


czardoz:
Transient issue with the CI, LGTM

PR #499, opened 2016-08-02
v3 => support for exporting globals, environment and collection

shamasis:
- separate export module
- support to export environment, collection, globals using the aforementioned export module
- JSON reporter which uses the export module
- README and BETA markdown updates

### DIFF
--- a/lib/cli/index.js
+++ b/lib/cli/index.js
@@ -115,6 +115,14 @@ var _ = require('lodash'),
             help: 'Specify a data file to use for iterations (either json or csv).'
         });
 
+        parser.addArgument(['-E', '--exportEnvironment'], {
+            help: 'Exports the environment to a file after completing the run'
+        });
+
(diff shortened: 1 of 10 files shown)


lib/cli/index.js
```diff
@@ -196,21 +196,33 @@ var _ = require('lodash'),
...
-        runParser.addArgument(['--reporters'], {
-            help: 'Specify the reporters to use for this run.',
-            defaultValue: 'cli'
-        });
-
         runParser.addArgument(['-d', '--iteration-data'], {
             help: 'Specify a data file to use for iterations (either json or csv)'
         });
 
+        runParser.addArgument(['--export-environment'], {
```
czardoz: Need to also map the legacy values to these ones

shamasis: Where? Just putting the command did not make a difference. 😕

PR #500, opened 2016-08-03
Moved all run related modules within ./lib/run folder

shamasis:


shamasis:
Merging as per discussion with @czardoz

PR #507, opened 2016-08-08
Added support for exit code suppression

kunagpal:


bin/newman.js
```diff
@@ -34,7 +34,7 @@ cli(process.argv.slice(2), 'newman', function (err, args) {
     dispatch(args, function (err) {
         if (err) {
             console.error(err);
-            process.exit(1);
+            process.exit(args.run.suppressExitCode ? 0 : 1);
```
shamasis: will this error out if `args.run` isn't present? In other words, is `args.run` guaranteed as object here (even for future where more commands will be in?

kunagpal: Yes, this will error out if `args.run` is absent.

bin/newman.js
```diff
@@ -34,7 +34,7 @@ cli(process.argv.slice(2), 'newman', function (err, args) {
     dispatch(args, function (err) {
         if (err) {
             console.error(err);
-            process.exit(1);
+            process.exit(_.toNumber(_.get(args, 'run.suppressExitCode', true)));
```
czardoz: Use a ternary operator in place of this is more suitable, because the behavior of _.isNumber is not documented if the argument is a boolean.
```
_.get(...) ? 1 : 0 
```
Also, should be false by default.

kunagpal: Currently, the behaviour of `lodash.toNumber` for `Boolean` values is identical to that of the `+` operator, so `true` is coerced to `1`, while `false` becomes `0`.

PR #521, opened 2016-08-09
Added exit code suppression capability

kunagpal:


npm/test-integration.js
```diff
@@ -61,8 +61,7 @@ module.exports = function (exit) {
                     collection: test.collectionJSON,
                     environment: test.environmentJSON,
                     globals: test.globalsJSON,
-                    iterationData: test.dataCSV || test.dataJSON,
-                    abortOnError: true
```
shamasis: If you don't send abort on error how will the Greg fail. It should have abortOnFailure

shamasis: Greg = test

czardoz: We need to have a hidden option here, that we can pass to the library. With stopOnFailure, we won't see any errors in the callback, because it's a graceful stop.

bin/newman.js
```diff
@@ -33,7 +33,7 @@ cli(process.argv.slice(2), 'newman', function (err, args) {
     dispatch(args, function (err) {
         if (err) {
             console.error(err);
-            process.exit(_.get(args, 'run.suppressExitCode') ? 1 : 0);
+            process.exit(_.toNumber(_.get(args, 'run.suppressExitCode', true)));
```
czardoz: 0:1

npm/test-integration.js
shamasis: This is integration test!

shamasis: Also - does stop on failure store the last error on which it stopped and pass it to summary.error?

czardoz: @kunagpal ^?

PR #535, opened 2016-08-13
automagically detect the length of the iteration data

### LINKED ISSUE #534: Dynamic iteration through data file no longer works?
Hi,

Previously I used v2 newman, and I'm sure when passing in data files, I used -n which iterated through all the rows. Using v3 now, it requires a number e.g. -n 2. Can we fix this?

Regards,

Aaron

czardoz:
Fixes #534


shamasis:
I like! Sadly all build failing! 🙄

kunagpal:
Not anymore.

lib/run/index.js
```diff
@@ -82,7 +82,8 @@ module.exports = function (options, callback) {
         runner.run(options.collection, {
             stopOnFailure: options.bail, // LOL, you just got trolled ¯\_(ツ)_/¯
             abortOnFailure: options.abortOnFailure,
-            iterationCount: options.iterationCount,
+            iterationCount: options.iterationCount ||
+                (_.get(options, 'iterationData') ? options.iterationData.length : 1),
```
shamasis: Will this not error out if iteration data is undefined or not an array? @czardoz

shamasis: Ignore above comment. 

Why not directly do _.get(options, "iterationData.length") || 1

czardoz: If iteration data is present, we should not pass any iteration count.

czardoz:
@kunagpal Can you add a small CLI level test? 

If we run Newman like this:
 
```
newman run collection --iteration-data ./my-csv.csv
```

By default, the number of iterations should equal the number of rows in the CSV.

lib/run/index.js
kunagpal: @czardoz So the intended order of precedence is:

1. iterationData.length
2. iterationCount
3. 1

kunagpal:
Alright.

lib/run/index.js
```diff
@@ -82,7 +82,7 @@ module.exports = function (options, callback) {
         runner.run(options.collection, {
             stopOnFailure: options.bail, // LOL, you just got trolled ¯\_(ツ)_/¯
             abortOnFailure: options.abortOnFailure,
-            iterationCount: options.iterationCount || _.get(options, 'iterationData.length') || 1,
+            iterationCount: _.get(options, 'iterationData.length') || options.iterationCount || 1,
```
czardoz: If iteration count is specified, it should override the length of iteration data.

test/unit/run.test.js
```diff
@@ -70,4 +71,58 @@ describe('run module', function () {
             }
         ], done);
     });
+
+    it('must correctly resolve conflicts between iterationData.length and iterationCount', function (done) {
```
czardoz: By CLI level tests, I meant these: https://github.com/postmanlabs/newman/blob/develop/test/cli/suppress-exit-code.test.js

You should add a spec like that, with iteration data as a CSV file, and then run it in these combinations:
1. `--iteration-data` absent and `--iteration-count` absent (should run one iteration)
2. `--iteration-data` present and `--iteration-count` absent (should run iterations equal to number of rows in iteration data)
3. `--iteration-data` absent and `--iteration-count` present (should run iterations equal to iteration count given)
4. `--iteration-data` present and `--iteration-count` present (should always run iterations equal to given iteration count)

lib/run/index.js
```diff
@@ -82,7 +82,7 @@ module.exports = function (options, callback) {
         runner.run(options.collection, {
             stopOnFailure: options.bail, // LOL, you just got trolled ¯\_(ツ)_/¯
             abortOnFailure: options.abortOnFailure,
-            iterationCount: options.iterationCount,
+            iterationCount: _.get(options, 'iterationData.length') || options.iterationCount || 1,
```
czardoz: iteration count should always override the length of iteration data

test/unit/run.test.js
kunagpal: Okay, how do we verify the number of iterations that occurred from within an `exec` call?

shamasis: By using nock to mock the API and on its reply add callback function with expect. 
https://www.npmjs.com/package/nock

But for now, simply add that it shouldn't throw non zero code and add a todo

lib/run/index.js
```diff
@@ -82,7 +82,7 @@ module.exports = function (options, callback) {
         runner.run(options.collection, {
             stopOnFailure: options.bail, // LOL, you just got trolled ¯\_(ツ)_/¯
             abortOnFailure: options.abortOnFailure,
-            iterationCount: options.iterationCount,
+            iterationCount: options.iterationCount || _.get(options, 'iterationData.length'),
```
shamasis: @czardoz just a thought - if runtime already handles this, we shouldn't even try parsing anything from iteration data.

Issue #537, opened 2016-08-16
empty globals file is generated when using --export-globals option


mykhaylo-:
Hi
My collection has the following code in one of the tests:
``` javascript
var token = postman.getResponseHeader("x-auth-token");
postman.setGlobalVariable("x_auth_token", token);
```
When I run the following command
```bash
newman run -r cli,html -e pub_env.json --bail --export-globals globals.json https://api.getpostman.com/collections/808581-9c4c8735-e0bc-cd88-16f7-b7600d75f416?apikey=$API_KEY
```
newman executes my collection successfully and generates globals.json file. But it is generated without any global variables while I expect "x_auth_token" to be there:
```javascript
{
  "values": [],
  "_postman_exported_at": "2016-08-15T20:53:08.375Z",
  "_postman_exported_using": "Newman/3.0.1"
}
```
Please suggest.

kunagpal:
@mykhaylo- Thanks for reporting this issue! A fix for this bug has been added within Newman `v3.1.0-beta.1`.

Please re-try exporting your run variables after installing Newman with `npm i -g newman@beta`, and let us know if your issue has been resolved.

mykhaylo-:
@kunagpal many thanks. it works in v3.1.0-beta.1

miguelC:
I am running into this issue with version 3.8.3 :(

PR #538, opened 2016-08-17
Fixed export bug

kunagpal:
In case of missing `environment` or `globals` options, a default value of `{}` is used as a fallback. This allows proper handling of variables created during a collection run.

Takes care of #537.

### DIFF
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -43,9 +43,16 @@ module.exports.get = (overrides, options, callback) => {
             return (src === null) ? dest : undefined;
         });
 
-
         if (_.isEmpty(loaders)) {
             return callback(null, options);
+        }
+        // sanitize environment option
+        if (!options.environment) {
+            options.environment = {};
+        }
+        // sanitize globals option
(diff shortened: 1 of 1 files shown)


czardoz:
Should merge after runtime is released

lib/run/index.js
```diff
@@ -71,6 +71,14 @@ module.exports = function (options, callback) {
         if (!_.isObject(options.collection)) {
             return callback(new Error('newman: expecting a collection to run'));
         }
+        // sanitize environment option
+        if (!options.environment) {
```
shamasis: Question - we do this here or in config module? Essentially config is supposed to sanitise it before it gets here

The fact that we are discussing this makes me think that we should organise it more evidently.

shamasis: I'm sorry but I think I wasn't that clear. I meant the config inside run. @czardoz can you look into it?

shamasis:
I'm sorry but I think I wasn't that clear. I meant the config inside run. @czardoz can you look into it?

PR #544, opened 2016-08-18
Lodash lint

shamasis:
+ added lodash eslint plugin
+ set sensible defaults in config (can be changed when issues around them is encountered)
+ also adds packity (which ensures that during testing, local packages match package.json)

Note that, since we prefer lodash for readability improvement in complex or highly nested code and only in modules where lodash is included to serve a utility function's purpose, we do not otherwise prefer lodash.

@kunagpal - should help you migrate to lodash v3 from v3 for other repositories


kunagpal:
L.G.T.M

Issue #332, opened 2016-02-19
Add summary to html report - Feature Request


MAOliver:
Just the same style of summary that is available in the text output would be fine.

Example:

> Summary:
>
> | Parent | Pass Count | FailCount |
> | --- | --- | --- |
> | Folder Owner | 164 | 72 |
> | Total | 164 | 72 |

shamasis:
This is WIP in v3. The current html template is crude and by 3.1 (hopefully) we will have a richer HTML

kunagpal:
@MAOliver The HTML template has been improved in Newman `v3.1.0`, do go ahead and try it out.

MAOliver:
@kunagpal very nice. Thanks for the attention.

PR #550, opened 2016-08-20
Improved HTML report template

kunagpal:
#Takes care of #546 and #332 
screen shot 2016-08-23 at 20 51 32


czardoz:
Can we attach the pass/fail counts to individual assertions rather than a request?

lib/reporters/html/index.js
```diff
@@ -1,18 +1,47 @@
...
+                result[currentExecution.id].failed += assertionStatus.failed || 0;
+            }, {});
+
+        _.forEach(aggregations, function (currentAggregation) {
+            _.set(currentAggregation, 'assertions', assertions[currentAggregation.item.id]);
+        });
+
+        // console.log(JSON.stringify(aggregations, null, 2));
+
+        _.set(this.summary.run, 'aggregations', aggregations);
```
czardoz: This should not be a part of the summary object

kunagpal: Yes, I'll adjust to accommodate individual assertion aggregations as well.

kunagpal:
Done.

kunagpal:
@shamasis Do take a look and let me know if there are any other tasks left on this.

lib/reporters/html/index.js
```diff
@@ -1,22 +1,89 @@
...
+            _.set(currentAggregation, 'assertions', _.values(_.get(assertions, currentId)));
+        });
+
+        _.set(this.summary.run, 'aggregations', aggregations);
+
         this.exports.push({
             name: 'html-reporter',
             default: 'newman-run-report.html',
-            path: options.export,
+            path: options.output,
```
czardoz: The rest of our reporters use `export`. Any reason for this?

kunagpal: `options.export` is no longer being used as a valid key, we had switched to output.

czardoz:
LGTM! @shamasis

PR #554, opened 2016-08-23
Modified the timestamp generator function to use a special separator …

shamasis:
Modified the timestamp generator function to use a special separator argument or use `-` as separator. This ensures that exported filenames are now more readable.

OLD: `newman-run-report-20160823093308656-0.html`
NEW: `newman-run-report-2016-08-23-10-40-03-637-0.html`


kunagpal:
@shamasis L.G.T.M

Issue #581, opened 2016-08-30
HTML Reporter


Mkoopajr:
So I'm not sure if this is a Windows issue or not. However I'm getting an error when running newman in the cli or as a node-module and using the html reporter. The cli and json reporters seem to be fine. The error is "TypeError: Expected a finite number" and seems to stem from the pretty-ms call [here](https://github.com/postmanlabs/newman/blob/develop/lib/reporters/html/index.js#L75).

lightswitch05:
Getting this on a mac too - so not windows specific

kunagpal:
@Mkoopajr @lightswitch05 The bug has been confirmed, successfully reproduced on our end, and fixed in #582. It should be added to the next Newman release.

PR #582, opened 2016-08-31
Fixed HTML reporter stats bug

kunagpal:
#581

### DIFF
--- a/lib/reporters/html/template-default.hbs
+++ b/lib/reporters/html/template-default.hbs
@@ -28,7 +28,7 @@
     <div class="row">
       {{#with summary}}
         <div class="col-md-3">Collection</div><div class="col-md-9">{{collection.name}}</div>
-        {{#if collection.description}}<div class="col-md-3">Description</div><div class="col-md-9">{{collection.description}}</div>{{/if}}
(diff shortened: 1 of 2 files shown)


lib/reporters/html/template-default.hbs
```diff
@@ -28,7 +28,7 @@
     <div class="row">
       {{#with summary}}
         <div class="col-md-3">Collection</div><div class="col-md-9">{{collection.name}}</div>
-        {{#if collection.description}}<div class="col-md-3">Description</div><div class="col-md-9">{{collection.description}}</div>{{/if}}
+        {{#if collection.description}}<div class="col-md-3">Description</div><div class="col-md-9">{{collection.description}}&nbsp;</div>{{/if}}
```
shamasis: Any reason? If it lacks description why would that row occupy space?

shamasis:
LGTM other than that one comment

lib/reporters/html/template-default.hbs
kunagpal: For empty collection descriptions, the row alignment is horribly skewed, such that the next line begins exactly where the description should have been.

shamasis: KGTMW

Issue #516, opened 2016-08-08
Windows build integration


kunagpal:
Use [Appveyor](https://www.appveyor.com/) to incorporate Windows build functionality.

shamasis:
added project

PR #584, opened 2016-09-01
Added Appveyor build configuration

kunagpal:
#516

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,12 +6,6 @@
 init:
   - git config --global core.autocrlf input
 
-# fetch repository as zip archive
-shallow_clone: true                 # default is "false"
-
-# set clone depth
-clone_depth: 5                      # clone entire repository history if not defined
-
 # environment variables
 environment:
   matrix:
@@ -19,17 +13,9 @@ environment:
     - nodejs_version: "5"
     - nodejs_version: "6"
 
-# build platform, i.e. x86, x64, Any CPU. This setting is optional.
-platform: Any CPU
-
-# build cache to preserve files/folders between builds
-cache:
-  - node_modules -> package.json                   # local npm modules
-
 # scripts that run after cloning repository
 install:
   - ps: Install-Product node $env:nodejs_version
-  - npm install -g npm
   - npm install
 
 build: off # Disable MSBuilds, not related to the regular build process
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,52 @@
+#---------------------------------#
+#    environment configuration    #
+#---------------------------------#
+
+# scripts that are called at very beginning, before repo cloning
+init:
+  - git config --global core.autocrlf input
+
+# fetch repository as zip archive
+shallow_clone: true                 # default is "false"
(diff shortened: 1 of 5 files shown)


appveyor.yml
```diff
@@ -0,0 +1,52 @@
+#---------------------------------#
+#    environment configuration    #
+#---------------------------------#
+
+# scripts that are called at very beginning, before repo cloning
+init:
+  - git config --global core.autocrlf input
+
+# fetch repository as zip archive
+shallow_clone: true                 # default is "false"
```
czardoz: @kunagpal What's the rationale behind this?

appveyor.yml
```diff
@@ -0,0 +1,52 @@
...
+
+# environment variables
+environment:
+  matrix:
+    - nodejs_version: "4"
+    - nodejs_version: "5"
+    - nodejs_version: "6"
+
+# build platform, i.e. x86, x64, Any CPU. This setting is optional.
+platform: Any CPU
```
czardoz: If optional, can we leave it out?

appveyor.yml
```diff
@@ -0,0 +1,52 @@
...
+
+# build cache to preserve files/folders between builds
+cache:
+  - node_modules -> package.json                   # local npm modules
```
czardoz: Can we disable caching between builds?
It makes failures very hard to debug

czardoz:
:cake:

appveyor.yml
kunagpal: It is optional, but this tests things more comprehensively.

appveyor.yml
kunagpal: That would slow down subsequent builds, but it's not a major issue.

test/cli/export-globals.test.js
```diff
@@ -1,20 +1,23 @@
...
     afterEach(function () {
-        try { fs.unlinkSync('out/test-globals.json'); }
+        try { fs.unlinkSync(exportedGlobalsPath); }
         catch (e) { console.error(e); }
     });
 
     it('`newman run` should export globals to a file', function (done) {
         // eslint-disable-next-line max-len
-        exec('./bin/newman.js run test/cli/single-get-request.json -g test/cli/simple-variables.json --export-globals out/test-globals.json', function (code) {
+        exec('node ./bin/newman.js run test/cli/single-get-request.json -g test/cli/simple-variables.json --export-globals out/test-globals.json', function (code) {
```
shamasis: How about making this change in exec itself?

kunagpal: Something like a persistent prefix?

shamasis: yes. Thoughts? Only downside is that if we run anything other than scripts, it will get mucked up.

Reason I say this is because, while writing tests, people will not remember and when Windows test fail, then they'll add back.

kunagpal: Given that we need the node context for all such CLI tests only, we can add it to the `test/cli` folder, with all other cases being the tester's responsibility.

PR #585, opened 2016-09-02
Added support and tests for --disable-unicode

kunagpal:
* Added capability to selectively pick between Unicode symbols and their plain text equivalents, depending on the detection of Unicode support / presense of `--disable-unicode`.
* Updated CLI options section in README.md and CLI parser to reflect the option updates
* Added tests to check for correct handling of Unicode detection and disabling.

**Note:** The build has failed due to an Appveyor configuration issue, and can be ignored for now.

### DIFF
--- a/lib/reporters/cli/index.js
+++ b/lib/reporters/cli/index.js
@@ -10,6 +10,7 @@ var _ = require('lodash'),
     LF = '\n',
     SPC = ' ',
     E = '',
+    symbols,
 
     /**
      * Helper function to get parent of an item
@@ -18,7 +19,7 @@ var _ = require('lodash'),
      * @returns {PostmanItemGroup}
      */
     parentOf = function (item) {
-        return item && item.__parent && item.__parent.__parent || undefined;
+        return item && item.__parent && item.__parent.__parent || undefined; // @todo: lodashify?
     },
 
     PostmanCLIReporter;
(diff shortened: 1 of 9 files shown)


czardoz:
@kunagpal Can you merge the latest `develop` into your branch? (that'll get AppVeyor to work properly).

kunagpal:
Yes, I'll have to edit tests as well.

shamasis:
I think, we should move these to a separate symbols module and then require it separately via CLI utils.

Currently,  whenever the symbols map function is called a new object is created. Instead we should try and have a map or let's say have a hash of the various types of symbols and return them accordingly that insures that we we do not create a new objects every time. Thoughts?

shamasis:
Also, the CLI tables need to comply. That needs a bit of discussion as to how it's done best.

lib/reporters/cli/index.js
```diff
@@ -18,7 +21,7 @@ var _ = require('lodash'),
      * @returns {PostmanItemGroup}
      */
     parentOf = function (item) {
-        return item && item.__parent && item.__parent.__parent || undefined;
+        return _.get(item, '__parent.__parent');
```
shamasis: this creates a string for every call!

kunagpal: Tuche.

PR #587, opened 2016-09-03
Update README instructions for JUnit reporting

groodt:


shamasis:
@groodt can you resolve the conflict and update the PR?

groodt:
I think the README was already updated in another branch. It is correct now on master.

PR #598, opened 2016-09-07
Added resolution check for slashed variable names

kunagpal:


kunagpal:
@czardoz @shamasis Is this good to go?

shamasis:
LGTM (with limited context)

PR #607, opened 2016-09-12
Restored `-v` to display Newman version

kunagpal:


lib/cli/index.js
```diff
@@ -237,10 +237,6 @@ var _ = require('lodash'),
             addHelp: false
         });
 
-        commonParser.addArgument(['-v', '--version'], {
```
shamasis: @czardoz were these duplicates? How!

lib/cli/index.js
```diff
@@ -384,13 +380,6 @@ var _ = require('lodash'),
             prog: program
         });
 
-        parser.addArgument(['--version'], {
```
shamasis: @czardoz were these duplicates? How!

shamasis:
LGTM. @czardoz - review and merge please.

lib/cli/index.js
```diff
@@ -384,13 +380,6 @@ var _ = require('lodash'),
...
-            help: 'Prints the version and exits',
```
shamasis: As per discussion with @czardoz we should keep the duplicates so that help text show up for legacy help text.

kunagpal: Okay.

PR #608, opened 2016-09-12
Added recursive test resolution capability

kunagpal:
* Re-organized integration test collections into respective individual folders
* Added recursive directory traversal in tests using `recursive-readdir`


shamasis:
Do we need to put all files in individual folders? I was thinking of putting all old integration tests into a `newman-v2-legacy` folder and rest stays in root and more than one related tests are put in folder.

kunagpal:
That tatters things a bit. Given the current structure, it makes more sens to put tests in their own folders, or none at all. Also, we haven't yet decided on what constitutes an old integration test.

npm/test-integration.js
```diff
@@ -51,7 +59,7 @@ module.exports = function (exit) {
 
             // run tests using the consolidated test set as a guide
             async.map(suites, function (test, next) {
-                console.info(` - ${test.name}`);
+                console.info(` - ${test.name.split('/')[2]}`);
```
shamasis: what's this @kunagpal ?

shamasis: what if I do not have a folder????

kunagpal: Interesting point. Tests have been arranged within folders, but that shouldn't be taken as a guarantee. I'll improve the logic here in the next commit.

kunagpal:
@shamasis That has already been handled, integration tests have been moved to folders only for those with multiple files under them.

PR #611, opened 2016-09-14
Removed iterationData inheritance from iterationCount

kunagpal:


czardoz:
@kunagpal How was this tested?

shamasis:
Looks good to me. Not sure if there were previous unit / integration tests watching out for this. Maybe a test should be there that send data file with X rows and it is tested that iteration runs for X rows. The other test is where data contains X rows but iterationCount is set to Y and test is done to check iteration runs for Y count.

czardoz:
@shamasis There were no tests watching for this, because we do not have a way (yet) to ensure that "n" iterations were run for CLI usage. 

@kunagpal How was this tested?

kunagpal:
@czardoz @shamasis: Testing manually for now, see #616

Issue #610, opened 2016-09-13
Tests not executing when using newman programmatically 


jonniedarko:
If I try to run my postman collection tests using the following code it does not fail but does not execute:
```js
var newman = require('newman'); // require newman in your project

newman.run({
    collection: require('./tests/api.postman_collection.json'),
    reporters: 'cli'
}, function (err) {
    if (err) { throw err; }
    console.log('collection run complete!');
});

```

this produces the output:
```bash
newman

api-server-pm-tests

┌─────────────────────────┬──────────┬──────────┐
│                         │ executed │   failed │
├─────────────────────────┼──────────┼──────────┤
│              iterations │        0 │        0 │
├─────────────────────────┼──────────┼──────────┤
│                requests │        0 │        0 │
├─────────────────────────┼──────────┼──────────┤
│            test-scripts │        0 │        0 │
├─────────────────────────┼──────────┼──────────┤
│      prerequest-scripts │        0 │        0 │
├─────────────────────────┼──────────┼──────────┤
│              assertions │        0 │        0 │
├─────────────────────────┴──────────┴──────────┤
│ total run duration: 4ms                       │
├───────────────────────────────────────────────┤
│ total data received: 0B (approx)              │
├───────────────────────────────────────────────┤
│ average response time: 0ms                    │
└───────────────────────────────────────────────┘
collection run complete!

```

But if I just run using the globally install newman command
```
newman run /pat/to/tests/api.postman_collection.json
```
I get the expected output:

```bash
newman

api-server-pm-tests

→ http://api.local.dev/xyz
  GET http://api.local.dev/xyz [200 OK, 385B, 750ms]
  ✓  Status code is 200
  ✓  No Error
  ✓  Correct Environment
  ✓  No Image

┌─────────────────────────┬──────────┬──────────┐
│                         │ executed │   failed │
├─────────────────────────┼──────────┼──────────┤
│              iterations │        1 │        0 │
├─────────────────────────┼──────────┼──────────┤
│                requests │        1 │        0 │
├─────────────────────────┼──────────┼──────────┤
│            test-scripts │        1 │        0 │
├─────────────────────────┼──────────┼──────────┤
│      prerequest-scripts │        0 │        0 │
├─────────────────────────┼──────────┼──────────┤
│              assertions │        4 │        0 │
├─────────────────────────┴──────────┴──────────┤
│ total run duration: 1093ms                    │
├───────────────────────────────────────────────┤
│ total data received: 91B (approx)             │
[...]

PR #613, opened 2016-09-15
Fixed V1 collection resolution bug

kunagpal:
Takes care of #610 

**Context:** Newman Library usage

This ensures that V1 collections passed as objects are unconditionally converted to their respective V2 equivalents first.

### DIFF
--- a/lib/run/options.js
+++ b/lib/run/options.js
@@ -57,35 +57,41 @@ var _ = require('lodash'),
      */
     configLoaders = {
         collection: function (value, callback) {
-            var done = function (err, collection) {
-                if (err) {
-                    return callback(err);
-                }
+            var V2_COLLECTION_PROPERTY = 'info.schema',
+                COLLECTION_TRANSFORMER_OPTION = { inputVersion: '1.0.0', outputVersion: '2.0.0' },
 
-                // ensure that the collection option is present before starting a run
-                if (!_.isObject(collection)) {
-                    return callback(new Error('newman: collection could not be loaded'));
-                }
+                done = function (err, collection) {
+                    if (err) {
+                        return callback(err);
+                    }
 
-                // ensure that the collection reference is an SDK instance
-                // @todo - should this be handled by config loaders?
-                collection = new Collection(Collection.isCollection(collection) ?
-                    // if the option contain an instance of collection, we simply clone it for future use
-                    // create a collection in case it is not one. user can send v2 JSON as a source and that will be
-                    // converted to a collection
-                    collection.toJSON() : collection);
+                    // ensure that the collection option is present before starting a run
+                    if (!_.isObject(collection)) {
+                        return callback(new Error('newman: collection could not be loaded'));
+                    }
 
-                callback(null, collection);
-            };
+                    // ensure that the collection reference is an SDK instance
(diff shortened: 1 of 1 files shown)


lib/run/options.js
```diff
@@ -57,35 +57,41 @@ var _ = require('lodash'),
      */
     configLoaders = {
         collection: function (value, callback) {
-            var done = function (err, collection) {
-                if (err) {
-                    return callback(err);
-                }
+            var V2_COLLECTION_PROPERTY = 'info.schema',
```
shamasis: Move this to top - outermost scope.

lib/run/options.js
```diff
@@ -57,35 +57,41 @@ var _ = require('lodash'),
...
+                        // create a collection in case it is not one. user can send v2 JSON as a source and that will be
+                        // converted to a collection
+                        collection.toJSON() : collection);
 
-            if (!_.isString(value)) { return done(null, value); }
+                    callback(null, collection);
+                };
+
+            if (_.isObject(value)) {
+                return _.get(value, V2_COLLECTION_PROPERTY) ? done(null, value) : transformer.convert(value,
```
shamasis: Would love to see some inline documentation

lib/run/options.js
```diff
@@ -57,35 +57,41 @@ var _ = require('lodash'),
...
+                        // if the option contain an instance of collection, we simply clone it for future use
+                        // create a collection in case it is not one. user can send v2 JSON as a source and that will be
+                        // converted to a collection
+                        collection.toJSON() : collection);
 
-            if (!_.isString(value)) { return done(null, value); }
+                    callback(null, collection);
+                };
+
+            if (_.isObject(value)) {
```
shamasis: I think this detection should explicitly check for v1 collection by looking for (requests array or folders array). Otherwise there will be a lot of false positives.

kunagpal: That is happening within this block. I can fine tune that inner condition but this is the handler for cases where collections are passed as objects.

lib/run/options.js
kunagpal: Will add.

lib/run/options.js
kunagpal: That's fine, but it's only being used within the collection load handler.

lib/run/options.js
shamasis: For now, move this to a separate function on top since this logic is used twice and if type === collection, run it. 

Tune the inner logic to treat v2 by default and transform to v1 if either folders exist or requests exist as array

lib/run/options.js
czardoz: @kunagpal The transformer now has functions called `isv1` and `isv2` to help out with this.

kunagpal: Cool.

lib/run/options.js
kunagpal: I'll be using the helper methods from `postman-collection-transformer` instead.

shamasis: check v1 else, assume v2 please.

kunagpal: Done

kunagpal:
@shamasis V1 collection detection has been made explicit, as discussed.

PR #614, opened 2016-09-15
Allowed run options to be specified as objects

kunagpal:
**Context:** Newman library usage

This ensures that data / environments / globals options specified as objects within `newman.run` are still processed correctly.


kunagpal:
@shamasis This is a reported issue on our Slack Community, in case you were wondering.

PR #627, opened 2016-09-22
Shifted print module to lib root

kunagpal:


lib/print/print.js
```diff
@@ -1,6 +1,6 @@
 
 var format = require('util').format,
-    cliUtils = require('./cli-utils'),
+    cliUtils = require('./../reporters/cli/cli-utils'),
```
czardoz: No need for `./` here. Just make it `../reporters/cli/cli-utils`

lib/reporters/cli/index.js
```diff
@@ -4,7 +4,7 @@ var _ = require('lodash'),
     format = require('util').format,
 
     cliUtils = require('./cli-utils'),
-    print = require('./print'),
+    print = require('./../../print/print'),
```
czardoz: Same comment as above

PR #637, opened 2016-09-26
Improved system tests

kunagpal:
* Removed `hashbangs` from `npm` scripts
* Removed system test for `hashbang` check in `npm` scripts
* Added check to ensure that npm scripts are of the form `/^node\snpm\/.+\.js$/`


shamasis:
@kumagpal - why? These scripts are also capable of independent execution.

kunagpal:
@shamasis They still are. Since they are access via `npm run`, the hashbang is effectively useless here.

czardoz:
@kunagpal By independant execution, @shamasis means on the command line, without `npm run ...`

PR #649, opened 2016-09-29
Fixed flaky sandbox integration test

kunagpal:
Randomly generated integers are now checked for the interval `>= 0`, instead of `> 1`.


shamasis:
This is okay to merge. But one observation is that we should test typeof and then parse int. since a lot of stuff can convert to 0

PR #672, opened 2016-10-07
Improved AppVeyor build configuration

kunagpal:


test/system/appveyor-yml.test.js
```diff
@@ -45,6 +45,12 @@ describe('appveyor.yml', function () {
             !travisYAMLError && expect(travisYAML.node_js).to.eql(appveyorNodeVersions);
         });
 
+        it('should have a valid install pipeline', function () {
+            expect(appveyorYAML.install[0].ps).to.be('Install-Product node $env:nodejs_version');
```
shamasis: These are becoming too strict a system check. Instead of verifying exact lines in array, it should have simply checked for the presence and relative position of the commands. Anyway, merging for now @kunagpal @czardoz

PR #678, opened 2016-10-09
Added in-code documentation for HTML, JUnit reporters

kunagpal:


lib/reporters/html/index.js
```diff
@@ -5,12 +5,38 @@ var fs = require('fs'),
     prettyMs = require('pretty-ms'),
     handlebars = require('handlebars'),
 
+    /**
+     * An object of the default file read preferences.
```
shamasis: These are `@private`

Issue #687, opened 2016-10-10
JUnit XML output does not conform to standard


mcdan:
The output emitted by the JUnit reporter is not correct. It is missing a required attribute on the testsuite node, "tests". This per the xsd available here: http://llg.cubic.org/docs/junit/.

mcdan:
This appears to be a regression around the refactor to v3 because other people report this working in v2.

shamasis:
The associated PR is merged and is out for release. Thanks everybody.

PR #688, opened 2016-10-10
Fix for #687 - add tests attribute to the junit reporter this

mcdan:
output correctly conforms to the XSD.

### DIFF
--- a/lib/reporters/junit/index.js
+++ b/lib/reporters/junit/index.js
@@ -73,12 +73,17 @@ JunitReporter = function (newman, reporterOptions) {
                         tests[name] = [];
                     }
                 });
+                if (execution.assertions) {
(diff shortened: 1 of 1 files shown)


mcdan:
I didn't build tests for this as there isn't a pattern for validating the output for the reporters that I could find in the test harness and I was hesitant to try and define it without further guidance.

shamasis:
@czardoz can you guide this to completion? And also possibly look as whether we can run all sanity tests with JUnit reporter and see how we can alleviate the XSD here http://llg.cubic.org/docs/junit/

@mcdan thank you 😍

czardoz:
Thanks @mcdan !

mcdan:
No problem. Ping me if you want help with the tests / a guinea pig for a
release.

PR #697, opened 2016-10-13
VariableScope support for export-import compatibility

shamasis:


shamasis:
merge only after runtime is bumped

shamasis:
Added more tests and bumped runtime. @czardoz - you should be free to review this now. :-)

shamasis:
@kunagpal - can you re run your export import tests with this branch and latest app?

jacobscarter:
Don't know if this helps at all, but I can verify that export from postman, manipulate in newman, and import into postman is working.

Using Postman v4.8.0 I exported an environment that had one variable needed to run the first collection in a set of 15 collections.  Here is an example of the 1 of the 15 newman commands I ran synchronously.

```
newman run ./Authentication.postman_collection.json -x -e ./jacob.postman_environment.json --export-environment ./jacob.postman_environment.json
```
Each test added variables that the next test needed.

When the tests were done I imported `jacob.postman_env....` into Postman App and it imported successfully and using Runner everything worked fine.

shamasis:
@jacobscarter - that was super helpful. We are so glad to have you help us with this. 😍

@czardoz - Time for a release on Monday then?

@kunagpal - we need to update our documentation if anywhere we used raw object as environment or globals, that will no longer be supported. Changelog should reflect that too.

PR #706, opened 2016-10-17
Improved .ignore system tests

kunagpal:


test/system/repository.test.js
```diff
@@ -154,27 +156,34 @@ describe('project repository', function () {
         });
     });
 
-    describe('.gitignore file', function () {
-        it('must exist', function (done) {
-            fs.stat('./.gitignore', done);
-        });
+    describe('.ignore files', function () {
+        var gitignorePath = './.gitignore',
```
czardoz: Surprisingly, looks like this works on windows.

kunagpal: This is not run via the command line, so yes.

PR #740, opened 2016-11-03
[WIP] Added after_success hook to .travis.yml

kunagpal:


kunagpal:
@shamasis @czardoz Requires documentation verification.

czardoz:
@kunagpal What all is pending in this PR?

kunagpal:
@czardoz Adding the keys as a part of the `.ci` folder, and a review of the existing JSDoc documentation.

shamasis:
Can we not add the key in Travis encrypted environment? Don't want to add it here at all.

kunagpal:
@shamasis How else could we do this? The pattern is the same as the collection-sdk.

shamasis:
Not sure if even that is correct. Keys should not be in repo usually. 

Shamasis Bhattacharya




On Fri, Nov 18, 2016 at 8:18 PM +0530, "Kunal Nagpal" <notifications@github.com> wrote:












@shamasis How else could we do this? The pattern is the same as the collection-sdk.

kunagpal:
Okay. Out of interest, let's explore some other means of achieving this.

czardoz:
Its not possible to set the keys as environment variables (there's a limit on the length). 

The best way is to encrypt the keys using travis cli and decrypt them at build time. @shamasis, there's no other better way of doing this. ⁣

kunagpal:
In that case, I'll add the keys in `. ci` and  within repository settings, while we go ahead with documentation review.

shamasis:
I'm good as long as @czardoz agrees that it is encrypted.

kunagpal:
@shamasis It is.

PR #749, opened 2016-11-10
Extra example for HBS template on options.reporter

eridem:
Hi!
When I was testing how to use a custom HBS template from the API reference, I was not sure if I could use `template` as attribute or not in the `options.reporter`. 
After testing, I figured out it was possible.
So, I was thinking that it could be a good idea add an extra example for it.


czardoz:
Thanks @eridem!

PR #772, opened 2016-11-22
Added more details to HTML report

kunagpal:
Partly inspired by #763 

* Shifted `prettyms` and `filesize` methods to `lib/util.js`
* Updated CLI and HTML reporters to use new util helpers.
* Added average and response times, total data transferred, and overall run duration to HTML report.
* Updated sample data in examples/reports

newman html report

PR #783, opened 2016-11-28
Removed superfluous details from summary response object

kunagpal:
https://github.com/postmanlabs/newman/pull/772#discussion_r89292749

### DIFF
--- a/lib/reporters/html/index.js
+++ b/lib/reporters/html/index.js
@@ -27,6 +27,13 @@ var fs = require('fs'),
      */
     ASSERTION_STATE = { false: 'passed', true: 'failed' },
 
+    /**
+     * The list of execution data fields that are aggregated over multiple requests for the collection run
+     *
+     * @type {String[]}
+     */
+    AGGREGATED_FIELDS = ['item', 'request', 'response'],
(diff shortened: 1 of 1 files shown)


kunagpal:
@czardoz That was the initial idea, but the `response.size` and `response.json` methods are used within the run instance, so I can't directly use the plain JSONified SDK object.

czardoz:
@kunagpal 
There's something specific about the response object that needs fixing. If this was a generic problem, we would have seen it in the `request` too.

kunagpal:
@czardoz Fixed.

lib/reporters/html/index.js
```diff
@@ -63,6 +63,8 @@ PostmanHTMLReporter = function (newman, options) {
                     // set base values for overall response size and time values
                     _.set(executionMeans, executionId, { time: { sum: 0, count: 0 }, size: { sum: 0, count: 0 } });
 
+                    currentExecution.response = currentExecution.response.toJSON();
```
czardoz: This mutates the executions array in `summary.run.executions`, and therefore is not an acceptable solution.

PR #794, opened 2016-12-02
Travis: Cache node_modules for faster builds

siddharthkp:
`npm install` takes ~24s (out of ~80s build time)

Travis added this feature back [in 2014](https://docs.travis-ci.com/user/caching/).

It will update `node_modules` only when `package.json` is modified.


czardoz:
@siddharthkp Thanks for this, but we want to not cache `node_modules`. The second, third & nth level dependencies need to be updated even if our own `package.json` has not changed.

Issue #796, opened 2016-12-04
Feature Request: Ability to get the values of Environment variables in html-reporter-template-data.json


Vijayabaskar03:
It would be great if there is a way get the values used for Environment Variables in the HTML report. This can also be used to pass on external values that are needed only for the sake of reporting.

hkgill:
+1

kunagpal:
@hkgill @Vijayabaskar03 This will be available in the next Newman release.

hkgill:
Thanks @kunagpal

kunagpal:
@hkgill @Vijayabaskar03 This is available in Newman `v3.3.1`, do go ahead and check it out :smile:

PR #801, opened 2016-12-07
Added globals, env vars to html report template data

kunagpal:
Inspired by #796


shamasis:
@kunagpal - we should add a test for the template object. In case we change any object structure of summary, the reporter will be broken and we will not even know.

PR #824, opened 2016-12-15
Added option to force colored output

kunagpal:
- [X] Added `--color` option.
- [X] Modified noTTY detection to work with the `--color` option.
- [X] Added documentation for `--color`, `options.color`.


czardoz:
Verified it works by piping the output to `cat` :cake:

Issue #825, opened 2016-12-16
HTML reporter templates should be able to access folder information for a request.


czardoz:
1. Newman Version (can be found via `newman -v`): latest
 2. OS details (type, version, and architecture): all
 3. Are you using Newman as a library, or via the CLI? both
 3. Did you encounter this recently, or has this bug always been there: It's a feature request
 4. Expected behaviour: The HTML reporter data should contain information about folders
 5. Command / script used to run Newman:
 6. Sample collection, and auxilliary files (minus the sensitive details):
 7. Screenshots (if applicable):

czardoz:
For example, instead of just "DigestAuth Request" here, we should also be able to indicate which folder this request is in. The folder information (basically just the name) should be a part of the template data.

![image](https://cloud.githubusercontent.com/assets/3726484/21256773/431e3e92-c39a-11e6-89ba-6e6872339900.png)

kunagpal:
:+1:

shamasis:
A path like approach is best like it is done in Newman failure table.

kunagpal:
Done in #826

kunagpal:
Pending on the `v3.4.0` release.

kunagpal:
Released with Newman `v3.4.0`.

PR #826, opened 2016-12-20
HTML report folder designation

kunagpal:
Inspired by #825 
- [X] Entire folder sections can be collapsed as needed.
- [X] Theoretically, folder mapping will work with infinitely deep folder nesting.
- [X] Works with folder only, singular requests only, and hybrid structured collections as well.

newman html report folders

### DIFF
--- a/lib/reporters/html/partials/aggregations.hbs
+++ b/lib/reporters/html/partials/aggregations.hbs
@@ -1,0 +1,56 @@
+{{#if folder}}
+    <div class="panel-group" id="collapse-folder-{{@index}}" role="tablist" aria-multiselectable="true">
+        <div class="panel-heading" role="tab" id="folderHead-{{@index}}">
+            <h4 class="panel-title"><a data-toggle="collapse" data-parent="#accordion" href="#folderData-{{@index}}" aria-controls="collapseOne"><strong>{{folder}}</strong></a></h4>
+        </div>
+
+        <div id="folderData-{{@index}}" class="panel-collapse collapse in" role="tabpanel" aria-labelledby="folderHead-{{@index}}">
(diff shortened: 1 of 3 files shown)


czardoz:
@kunagpal I think it might be better to have folder names in the titlebar of each request, separated by a `/`. It can also become very repetitive though. 

@shamasis Thoughts?

lib/reporters/html/partials/aggregations.hbs
```diff
@@ -0,0 +1,56 @@
+{{#if folder}}
```
czardoz: While I know that it's generally a good practice to separate templates out into partials, it's unnecessary in this case, because it is being used only in one place. 

@kunagpal Unless there's any other reason you've made this change, I think we should keep the template to a single handlebars file.

kunagpal: @czardoz Partialization has been done in order to account for deeply nested folder structures. The same structure can effectively be recursively re-used to make folder divisions.

As for folder name placement, keeping them separated from the request titlebar makes it possible to collapse entire folders, something that would be cumbersome with the `request / folder` approach.

malaubay: If the html report template is separated into 2 ".hbs" files, is it possible to customize both the "template-default.hbs" and "aggregations.hbs", and define the templates used in command line call (e.g. --reporter-html-template <path>)?

czardoz: @malaubay Can you open a new issue for your query? It'll be easier for other people who have the same question to find the answer :)

malaubay: Thanks, #893 reported.

PR #854, opened 2017-01-05
Fixes #852

### LINKED ISSUE #852: HTML Reporter Issue from 3.4.0
This error manifests itself when using the HTML Reporter and an uncaught error is thrown in request test.  Forgive me for not pulling and fixing myself as I have never contributed to a github project before!

 1. Newman Version : **3.4.0**
 2. OS details: **Windows 10 x64**
 3. Are you using Newman as a library, or via the CLI? **CLI**
 3. Did you encounter this recently, or has this bug always been there: **Recently**
 4. Expected behaviour: **Test run finishes and display error in output/html**
`newman run HtmlReporterIssue.postman_collection.json --reporters html,cli --bail`
**(see attachment)**
[HtmlReporterIssue.postman_collection.zip](https://github.com/postmanlabs/newman/files/686024/HtmlReporterIssue.postman_collection.zip)

2017-01-04 14_36_46-command prompt

(This can be reproduced by running the supplied collection with the command above.

SETUP
1. Have more than 1 request in a collection.
2. Have the 1st collection throw an uncaught exception in its test, in this case:
> var responseJSON;
>     responseJSON = undefined; 
>     tests["Status equals 200"] = responseJSON.status === 200; //error thrown here

Execution
1. Run Collection with newman using command above, noting the use of html reporter and bail flags
2. Receive error in screenshot

----------------------------------------
Analysis
The issue seems to be from line 113 in the \html\index.js file:
`var aggregationMean = executionMeans[item.id]`
Where `executionMeans` has an undefined entry for the **second** request in the collection, as it was never run...possibly due to the --bail flag added to the command.  Perhaps a sanity null check of aggregationMean before continuing to retrieve data on this object before continuing might suffice.

------------------------------------------
Workarounds
1. try/catch blocks around offending code in request test
    **Not ideal as some tests might not contain this logic**
2. not executing the collection with --bail
   **Even less ideal in my case as I am attempting to attach the collection runs to a CI process which should fail the build if any errors arise on execution.**

--Nathan

kunagpal:


test/cli/html-reporter.test.js
```diff
@@ -0,0 +1,49 @@
+var fs = require('fs');
+
+describe('HTML reporter', function () {
```
czardoz: These tests don't actually test the generated files, so I think we should name the file accordingly. `shallow-html-reporter.test.js` maybe? @kunagpal

PR #862, opened 2017-01-09
corrected the changelog message :@

czardoz:

PR #917, opened 2017-02-16
Converting JUnit Reporter responseTime

bteng22:
Resolves issue #862 Converting JUnit reporter responseTime from milliseconds to seconds


bteng22:
@kunagpal Anytime! Thanks for the quick response :D

PR #943, opened 2017-03-08
Resilient csv parsing

kunagpal:


czardoz:
What was the original behaviour?

kunagpal:
@shamasis @czardoz The default escape character is `"`, which is has now been changed to `\`.

shamasis:
Seems in the right direction. But this could be breaking change for many. And as such, should be a minor upgrade candidate with enough caveat. 

Any way we can test the impact on a greater subset?

shamasis:
However if the app already behaves like this, we can go ahead and merge.

kunagpal:
@shamasis As @madebysid has pointed out, the app does behave like this.

PR #994, opened 2017-04-18
Upgrade to second generation sandbox 

shamasis:
- added support in CLI reporter for skipped assertions
- added listener to runtime assertion event and bubbled up the same (had to tweak the assertion counting logic to make it usable)


lib/reporters/cli/index.js
```diff
@@ -307,11 +313,6 @@ _.assignIn(PostmanCLIReporter, {
             failure.source &&
                 (message += format(colors.gray('\ninside "%s"'), util.getFullName(failure.source)));
 
-            // augment message with item parent information
```
czardoz: Because the fullname is being printed in line 308

PR #996, opened 2017-04-20
Improved logging for deeply nested objects/long arrays

kunagpal:
```json
{
    "depth": 25,
    "maxArrayLength": 100
}
```


czardoz:
@BigstickCarpet @shamasis Ping ^

shamasis:
Usually, people will not do console.log unless they're debugging. And there are flags to turn off console output. Plus we can make a dedicated console-only reporter!

So, let's change the default depths for now. If anyone grumbles, we can give CLI option to control these values.

PR #1011, opened 2017-05-02
Add response body to CLI reporter

mbaeuerle:
When doing automated testing with newman we often encounter the problem, that a test is failing but we don’t know why exactly. So here the response body is very helpful in giving a hint to what really went wrong.
I wanted to include the body only when a test script is failing but unfortunately this information is not passed in the function I changed.

### DIFF
--- a/lib/reporters/cli/index.js
+++ b/lib/reporters/cli/index.js
@@ -119,8 +119,8 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
         size = size && (size.header || 0) + (size.body || 0) || 0;
 
         err ? print.lf(colors.red('[errored]')) :
-            print.lf(colors.gray('[%d %s, %s, %s]'), o.response.code, o.response.reason(),
-                util.filesize(size), util.prettyms(o.response.responseTime));
+            print.lf(colors.gray('[%d %s, %s, %s]\nResponse body:\n%s'), o.response.code, o.response.reason(),
+                util.filesize(size), util.prettyms(o.response.responseTime), o.response.body);
     });
 
     // Print script errors in real time


kunagpal:
Hey @mbaeuerle, thanks for your contribution. Unfortunately, this method of adding information to reporters for specific use cases (such as this one) is somewhat anti-pattern. However, we encourage you to [create and use your own custom Newman reporter](https://github.com/postmanlabs/newman#creating-and-using-custom-reporters) for the same :smile:

mbaeuerle:
@kunagpal we tried to write a custom reporter based on the internal CLI one and just add the response body. But it turns out the CLI reporter is written in a way which is not working when using it as an external reporter.

Let me explain our use case of newman:
- We use newman in our CI cycle to check if we have any regressions in our API 
- We have several Postman collections each testing a different use case
- These use cases can be rather long (~ 10 to 30 requests)
- Now if one of these requests fail with say a `422` we don't have a clue on what went wrong, because the error message is in the response body which is not shown
- To reproduce the error message we have to step through every request by hand until the one which is broken (we can't use the Postman Runner for this, because it only stores the body of 10 requests)

I think this and similar scenarios are not uncommon.
So I have another suggestion: 
Introduce a new newman option say `--show-response-body-on-failure`. When enabled the body of the failing request is shown. 
If this is acceptable I will create another PR with this behavior.

kunagpal:
@mbaeuerle Could you point me to the code for the custom reporter you created?

mbaeuerle:
We now created a working reporter (https://github.com/semlabs/newman-reporter-phpunit). It works in newman `3.5.2' but unfortunately it doesn't work with newman `3.6.0` because it seems the response body is not even passed to the reporters any longer.

kunagpal:
@mbaeuerle The string `body` property has been replaced by a buffer: `stream`. See the changes above. ☝️

mbaeuerle:
@kunagpal thank you very much!

shamasis:
If you do response.text() you should get the string converted from buffer.

shamasis:
The objects are documented roughly at - http://www.postmanlabs.com/postman-collection/

PR #1039, opened 2017-05-23
Update JUnit reporter to add time attribute to testsuites and testcas…

dopsonbr:
…e elements.

The time attribute is required for the testsuites and testcase elements per the
JUnit spec.


lib/reporters/junit/index.js
```diff
@@ -17,8 +17,12 @@ JunitReporter = function (newman, reporterOptions) {
         var report = _.get(newman, 'summary.run.executions'),
             collection = _.get(newman, 'summary.collection'),
             cache,
-            root;
+            root,
+            testSuitesExecutionTime,
+            executionTime;
 
+        testSuitesExecutionTime = 0;
```
kunagpal: Minor change: The initializations in these two lines can be shifted above.

dopsonbr: Thanks. Ill make the changes and merge develop into this branch today

dopsonbr:
@shamasis I agree that the test suites execution time should be the total execution not just the summation of the response times. However, it appears the current code is setting the test suite time to the response time and the test suites execution time should be the summation of each test suite.

I wanted to make the changes as small as possible because this is my first open source contribution and am still learning the rules of the road.

lib/reporters/junit/index.js
dopsonbr: The changes have been made and branch updated.

kunagpal:
This change is now available in Newman v3.6.1, do check it out 😄

PR #1056, opened 2017-06-06
Test

Malgalad:


Malgalad:
My bad, accidentally opened PR when intending to to do the reverse.

PR #1058, opened 2017-06-08
CSV parse improvement

### LINKED ISSUE #1054: Newman Cli Treats Number as String in CSV Data Files
1. Newman Version (can be found via `newman -v`): **3.6.0**
 2. OS details (type, version, and architecture): **Windows 10 Pro (x64)**
 3. Are you using Newman as a library, or via the CLI? **Cli**
 3. Did you encounter this recently, or has this bug always been there: It's not necessary a bug. But it's been at least since v3.4.3
 4. Expected behaviour: 
When running a collection with iterationData (**CSV**), the **number** type should be treated as **number** type. 

I debugged it. Newman is using **csv-parse**. Here is the code

In **newman\lib\run\options.js** 
Line: 217

`parseCsv(data, { columns: true, escape: '\\' }, cb);`

If I change it to 
`parseCsv(data, { columns: true, escape: '\\', auto_parse: true }, cb);`

Then the number would be treated as a number, and the test would pass.

 5. Command / script used to run Newman: **newman run "./CSV Data Number Type.postman_collection.json" -d "./SampleData.csv"**
**CSV Data Number Type.postman_collection.json**
`{
	"variables": [],
	"info": {
		"name": "CSV Data Number Type",
		"_postman_id": "ca1eff26-60a2-98a1-a229-bb7a0a8bc5a8",
		"description": "",
		"schema": "https://schema.getpostman.com/json/collection/v2.0.0/collection.json"
	},
	"item": [
		{
			"name": "Load Csv Columns As Number",
			"event": [
				{
					"listen": "test",
					"script": {
						"type": "text/javascript",
						"exec": [
							"tests[\"NumberColumn is \" + typeof(data.NumberColumn) + \" type \"] = true ;",
							"",
							"tests[\"NumberColumn \" + data.NumberColumn + \" should be 1 \"] = data.NumberColumn === 1;"
						]
					}
				}
			],
			"request": {
				"url": "http://www.google.com",
				"method": "GET",
				"header": [],
				"body": {},
				"description": ""
			},
			"response": []
		}
	]
}`

**SampleData.csv**
NumberColumn,StringColumn
1,hi

1. Create 2 files **CSV Data Number Type.postman_collection.json** and **SampleData.csv** using the code above
2. **newman run "./CSV Data Number Type.postman_collection.json" -d "./SampleData.csv"**
3. One test fails, as the type of data.NumberColumn is **string**.

Suggestions:
When I run it using the **Postman runner**, it's all good. When I use **JSON** as the data file format, it's also ok.
I'm not sure is it possible to provide a way to treat a number as number type. Probably the same way as Postman runner handles the number type.

kunagpal:
Fixed #1054


czardoz:
@kunagpal Check `is_int` so that we don't auto parse anything other than a number

PR #1098, opened 2017-06-28
Restore HTML reporter response body

kunagpal:
This does not mutate the original execution summary


cchretien:
This is good. While you're in the area...can you expose all the various errors to html reporter?
I.e. Request Error types and Script Reference errors? Currently, all that is exposed is counts for these type of errors.

kunagpal:
@cchretien While your suggestion is beyond the scope of the inbuilt HTML reporter, feel free to create your own custom reporter: https://www.getpostman.com/docs/postman/collection_runs/command_line_integration_with_newman#custom-reporters 😄

Issue #1037, opened 2017-05-22
v3.6.0 no longer has `summary.run.executions.response.body`


deyhle:
1. Newman Version (can be found via `newman -v`): 3.6.0
 2. OS details (type, version, and architecture): macOS 10.12.5 / node.js v6.10.2
 3. Are you using Newman as a library, or via the CLI? Library
 3. Did you encounter this recently, or has this bug always been there: since 3.6.0
 4. Expected behaviour: 
Until the latest release (up to 3.5.*) the `summary.run.executions` objects would include `response` with a property `body`. This was the response body as string. Since the Update to 3.6.0 the `body` property (`summary.run.executions[0].response.body`) is undefined. There seems to be no way to access the response body now?

kunagpal:
@deyhle The `v3.6.0` equivalent is `summary.run.executions[0].response.text()`.

deyhle:
Thanks for the quick response. 
However, `summary.run.executions[0].response.text()` also returns undefined, `summary.run.executions[0].response.json()` (the response is JSON) fails with the following error:

> JSONError: Unexpected token u in JSON at position 0
    at Object.parse (native)
    at parse (.../node_modules/liquid-json/lib/index.js:36:21)
    at Object.parse (.../node_modules/liquid-json/lib/index.js:58:20)
    at Object.json (.../node_modules/postman-collection/lib/collection/response.js:276:22)

(apparently it tries to `JSON.parse` undefined)

When downgrading to newman v3.5.2 it works and the body parameter is there. 
I just JSON.parsed the `summary.run.executions[0].response.body` and everything was fine. It appears that somewhere in the current version the body is swallowed, when even the accessor functions `text()` and `json()` can't read it.

kunagpal:
@deyhle Could you share the collection - environment pair that causes this? Also, could you see if `summary.run.executions[0].response.stream.toString()` works correctly?

deyhle:
Alright, here is my simplified code.
`summary.run.executions[0].response.stream` is also undefined.

```js
const newman = require('newman');

const collection = {
  "id": "020f340c-6dc8-0c9a-7ea9-76dde03b6a44",
  "name": "postmantest",
  "description": "",
  "order": [
    "0aae62d7-1ea7-0f85-31bf-3286e609206d"
  ],
  "folders": [],
  "timestamp": 1495454706474,
  "owner": "91456",
  "public": false,
  "requests": [
    {
[...]

PR #1099, opened 2017-06-29
Update index.js

aksbenz:
converted response.body to buffer otherwise toString of standard object is called returning [object Object] as a string. Refer to https://github.com/postmanlabs/newman/issues/1037

### DIFF
--- a/lib/reporters/html/index.js
+++ b/lib/reporters/html/index.js
@@ -76,7 +76,7 @@ PostmanHTMLReporter = function (newman, options) {
 
                     if (reducedExecution.response && _.isFunction(reducedExecution.response.toJSON)) {
                         reducedExecution.response = reducedExecution.response.toJSON();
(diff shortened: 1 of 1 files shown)


malaubay:
reducedExecution.response.body = Buffer.from(reducedExecution.response.stream).toString();
newman returned error on this line: "TypeError: this is not a typed array".

aksbenz:
Whats the node version ?
Check this : [Getting TypeError: this is not a typed array using Buffer.from](https://stackoverflow.com/questions/36899888/getting-typeerror-this-is-not-a-typed-array-using-buffer-from-in-mocha)

aksbenz:
@czardoz @kunagpal 
Buffer.from is available in node v4.5, 5.10 and 6 onwards.
To support any other version will need to change this ?
`reducedExecution.response.body = (new Buffer(reducedExecution.response.stream)).toString(); `

kunagpal:
@askbenz See #1105

PR #1103, opened 2017-07-03
Upgrade nodejs version to 6

shankyty:
Upgrade nodejs version to 6, for removing error Buffer.from
TypeError: this is not a typed array.
    at Function.from (native)
    at /home/shashanktyagi/repo/regression/node_modules/newman/lib/reporters/html/index.js:79:65
    at /home/shashanktyagi/repo/regression/node_modules/newman/node_modules/lodash/lodash.js:13767:16
    at arrayEach (/home/shashanktyagi/repo/regression/node_modules/newman/node_modules/lodash/lodash.js:537:11)
    at Function.transform (/home/shashanktyagi/repo/regression/node_modules/newman/node_modules/lodash/lodash.js:13766:43)
    at EventEmitter.<anonymous> (/home/shashanktyagi/repo/regression/node_modules/newman/lib/reporters/html/index.js:60:28)
    at EventEmitter.emit (/home/shashanktyagi/repo/regression/node_modules/newman/node_modules/eventemitter3/index.js:151:33)
    at _.assignIn.done (/home/shashanktyagi/repo/regression/node_modules/newman/lib/run/index.js:221:29)
    at /home/shashanktyagi/repo/regression/node_modules/newman/node_modules/postman-runtime/lib/backpack/index.js:56:34
    at PostmanCollectionRun._.assign._process (/home/shashanktyagi/repo/regression/node_modules/newman/node_modules/postman-runtime/lib/runner/run.js:133:20)

PR #1105, opened 2017-07-04
Fixed Buffer construction issues on Node v4

### LINKED ISSUE #1102: Error: this is not a typed array
[This change is introduced in version 6](https://github.com/postmanlabs/newman/commit/f86a9efb92ef92dcb058fe93995861452f25c1d5)

Please refer [this](https://nodejs.org/api/buffer.html#buffer_buffer)

Hence need to upgrade node version in package.json

kunagpal:
Resolves #1102, closes #1103

### DIFF
--- a/lib/reporters/html/index.js
+++ b/lib/reporters/html/index.js
@@ -34,7 +34,7 @@ var fs = require('fs'),
      */
     AGGREGATED_FIELDS = ['item', 'request', 'response'],
 
-    IS_NODE4 = process.version.match(/^v(\d+)\./)[1] === '4',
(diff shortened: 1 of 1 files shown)


lib/reporters/html/index.js
```diff
@@ -34,6 +34,8 @@ var fs = require('fs'),
      */
     AGGREGATED_FIELDS = ['item', 'request', 'response'],
 
+    IS_NODE4 = process.version.match(/^v(\d+)\./)[1] === '4',
```
shamasis: This needs protection.

kunagpal: Added protection

PR #1115, opened 2017-07-06
Ensure that failed fetches from the Pro API result in a non zero exit code

### LINKED ISSUE #1055: Newman run should not return success with a not found collection
**I believe this bug can have huge consequence with an continuous integration system!!!**

 1. Newman Version (can be found via `newman -v`): 3.6.0
 2. OS details (type, version, and architecture): docker & osx
 3. Are you using Newman as a library, or via the CLI? cli
 4. Expected behaviour:
`newman run https://www.getpostman.com/collections/651de8da5b5f-not-found`
screen shot 2017-06-03 at 10 27 38

`newman run https://www.getpostman.com/collections/651de8da5b5f-not-found`
`echo $?` shows `0`

kunagpal:
Fixes #1055

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -19,6 +19,7 @@ var fs = require('fs'),
      */
     FILESIZE_OPTIONS = { spacer: '' },
 
+    PRO_API_HOST = 'api.getpostman.com',
     USER_AGENT_VALUE = 'Newman/' + version;
 
 util = {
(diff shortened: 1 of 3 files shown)


lib/util.js
```diff
@@ -103,6 +104,19 @@ util = {
                     return callback(_.set(e, 'help', `the url "${location}" did not provide valid JSON data`));
                 }
 
+                var error,
+                    resource;
+
+                if ((response.statusCode !== 200) || (body && body.error)) {
```
shamasis: only stop on non 200 code but, if body.error is present, use it only to extract message.

lib/util.js
```diff
@@ -103,6 +104,19 @@ util = {
...
+                    error = new Error(_.get(body, 'error.message'));
```
shamasis: If body was not containing `error`, then error will be undefined.

kunagpal: Addressed

lib/util.js
kunagpal: Addressed

PR #1160, opened 2017-08-07
IO event support

kunagpal:


lib/run/index.js
```diff
@@ -26,6 +26,7 @@ var fs = require('fs'),
         request: ['response', 'request', 'item', 'cookies'],
         beforeTest: ['events', 'item'],
         test: ['executions', 'item'],
+        io: ['trace', 'response', 'request', 'cookies'],
```
czardoz: Arguments depend on the type of IO event.. check what "trace" contains. A request, response and cookies only make sense for HTTP IO.

shamasis: We might have to rethink `io` event. This variadic property is causing a pain for all consumers. App as well as newman.

shamasis:
We need to revisit how IO event looks like before we do this. @kamalaknn

PR #1164, opened 2017-08-11
The html reporter generate the same block for different iteration.

HOOHH:
Since we run the newman for testing the APIs,we need to run a collection in difference parameter like:
https://github.com/auth=1
https://github.com/auth=2
Make a CSV file to let it run in a collection,and actually the html report return 
https://github.com/auth=1
https://github.com/auth=1
results in report.


kunagpal:
@HOOHH Thanks for your contribution, but the HTML reporter is intended to aggregate request statistics across multiple iterations by design. However, you can publish this change as a [custom Newman reporter](https://www.getpostman.com/docs/postman/collection_runs/command_line_integration_with_newman#custom-reporters), which will let you use this customization (any other desired ones) as well 😄

HOOHH:
thanks

Issue #1203, opened 2017-09-15
feature request to turn off text wrapping or force a width larger than 60 for console.log events


cchretien:
The CLI reporter should have to ability to configure the text wrap limit for console.log events.
In a build environment such as TeamCity we are restricted to just [60 chars](https://github.com/postmanlabs/newman/blob/develop/lib/reporters/cli/cli-utils.js#L117). The requests api's and their stats are not wrapped at all. This behaviour was not present in release 2

The user should be able to....
* to turn off wrap for console.log events
* and/or, increase the default width from [60](https://github.com/postmanlabs/newman/blob/develop/lib/reporters/cli/cli-utils.js#L117)

Version and environment information:
 1. Newman Version (can be found via `newman -v`): 3.5.2
 2. OS details (type, version, and architecture): All Supported
 3. Are you using Newman as a library, or via the CLI? Library
 3. Did you encounter this recently, or has this bug always been there: encountered in release 3. Not present in release 2
 4. Expected behaviour: ability to turn off wrap or increase text width that is defaulted to 60 for console.log events
 5. Command / script used to run Newman: N/A
 6. Sample collection, and auxilliary files (minus the sensitive details):
 7. Screenshots (if applicable):

Steps to reproduce the problem:
1. engineer a collection, where in the test script, there is console.log instruction that prints out a string longer than 60 chars
2. run newman library in an environment where the shell is not a tty. (i.e. in a build environment such as TeamCity)
2. the wrap will default to 60

**Note:** I've tried setting `TERM=xterm` in the environment or `CI=false`

shamasis:
We could give --terminal-width as an option.

timostark:
Hi,

Do you have any update here? I don't see a smart way to avoid the line breaks in jenkins. The current setup (line break every 60 characters), makes the output super ugly :-(

Regards,
Timo

ChieftainY2k:
Hello, any news on that ? The line wrap limit at 60 is reeaaaaal PITA , if you will :)

imagest:
Would really like this as well.

thedomdom:
+1 Me too :)

cyansprite:
Would love to see this as well

felitra:
No updates on this one?

anil-shekhar-sage:
https://github.com/postmanlabs/newman/pull/492/files

piorus:
+1

robstoll:
[...]

PR #1230, opened 2017-10-04
Adjustable terminal width

kunagpal:
This takes care of #1203


kunagpal:
Needs to be reworked for `commander`.

connorsadlervelo:
Is this not supported in the newer versions of newman?
I am seeing wide console logging from the command line in Mac OS but it's limited to 80 chars when run in Jenkins

kunagpal:
@connorsadlervelo This was to allow tweaking the width restriction applied to console output within collection scripts. Regular CLI reporter output will try to intelligently detect the current width, as can be seen here: https://github.com/postmanlabs/newman/blob/develop/lib/reporters/cli/cli-utils.js#L133

connorsadlervelo:
Thanks for that - should that intelligent logic work on Jenkins? It seems to limit the width there

shamasis:
@kunagpal also when we rework on this, the option should be CLI reporter specific

PR #1262, opened 2017-11-04
Migrated Newman from argparse to commander for CLI argument parsing.

deepakpathania:
- Moved from argparse to commander for CLI argument parsing.
- Removed legacy options and their tests which were deprecated.
- Moved the entire CLI parsing logic to `bin/newman.js` and deleted `lib/cli/index.js`
- Added and modified tests for commander specific functionality.


deepakpathania:
Integration tests are failing which are unrelated to the changes made.

kamalaknn:
@deepakpathania the fix for this is merged on upstream. Updating your branch against upstream should fix the tests.

deepakpathania:
@kamalaknn Hey, thanks for the update, adding a few more functionalities, would update the branch against upstream once that is done.

bin/newman.js
```diff
@@ -2,11 +2,261 @@
...
+        };
+    },
+
+    /**
+     *  used for collecting global key=value variables supplied through command line
+     *
+     * --global-var "foo=bar" --global-var "alpha=beta"
+     *
+     * @param {String} val - The argument provided to `--global-var`.
+     * @param {String} memo - The array that is populated by key value pairs.
```
harryi3t: `@param {Array} memo` ?

bin/newman.js
```diff
@@ -2,11 +2,261 @@
...
+     * @returns {{key, value}} - The object representation of the current CLI variable.
+    */
+    collect = function (val, memo) {
+        var arg = val,
+            eqIndex = arg.indexOf('='),
```
harryi3t: we can avoid changing the datatype of `arg` by
```js
var arg,
   eqIndex = val.indexOf('='),
   arg = ... ;
```

deepakpathania:
@harryi3t Hey, I have made a separate PR #1278 with certain changes. Can you please review that instead? Closing this one.

bin/newman.js
```diff
@@ -2,11 +2,261 @@
...
+    },
+
+    /**
+     *  used for collecting global key=value variables supplied through command line
+     *
+     * --global-var "foo=bar" --global-var "alpha=beta"
+     *
+     * @param {String} val - The argument provided to `--global-var`.
+     * @param {String} memo - The array that is populated by key value pairs.
+     * @returns {{key, value}} - The object representation of the current CLI variable.
```
harryi3t: This returns an array right not object?

deepakpathania: Yeah, it should be `[{key,value}]`.

bin/newman.js
deepakpathania: Yes, that entire comment needs changes.

bin/newman.js
deepakpathania: Noted.

PR #1271, opened 2017-11-07
Update postman-collection@3.0.3 and postman-runtime@7.0.1

kamalaknn:
close #1265 

- [x] Incorporate global timeout option properly, and set default to 0
- [x] Should legacy test assertions be sent with `error` from `postman-sandbox`?


lib/run/index.js
```diff
@@ -177,26 +182,29 @@ module.exports = function (options, callback) {
...
-                            '   at Object.eval sandbox-script.js:' + (assertion.index + 1) + ':' +
-                            ((cursor && cursor.position || 0) + 1) + ')'
-                    }), {
-                        cursor: cursor,
-                        assertion: assertion.name,
-                        skipped: assertion.skipped,
-                        error: assertion.error,
-                        item: run.resolveCursor(cursor)
+                        emitter.emit('assertion', (assertion.passed ? null : {
+                            name: _.get(assertion, 'error.name', 'AssertionFailure'),
```
kamalaknn: @shamasis  Legacy tests do not have `error` on the Assertion. I think `postman-sandbox` should set an `AssertionFailure` error for legacy errors.

lib/run/index.js
```diff
@@ -250,32 +258,9 @@ module.exports = function (options, callback) {
...
-
-                        stack: 'AssertionFailure: Expected tests["' + assertion + '"] to be true-like\n' +
-                            '   at Object.eval ' + type + '.js:' + (index + 1) + ':' +
-                            ((cursor.position || 0) + 1) + ')'
-                    }), _.assignIn({ assertion: assertion }, o));
-                });
-
                 // bubble special script name based events
-                type && emitter.emit(type + 'Script', err, o);
+                o && o.event && o.event && emitter.emit(o.event.listen + 'Script', err, o);
```
harryi3t: `o.event` is tested twice

lib/run/index.js
```diff
@@ -177,26 +180,29 @@ module.exports = function (options, callback) {
...
-                        assertion: assertion.name,
-                        skipped: assertion.skipped,
-                        error: assertion.error,
-                        item: run.resolveCursor(cursor)
+                        emitter.emit('assertion', (assertion.passed ? null : {
+                            name: _.get(assertion, 'error.name', ''),
+                            index: assertion.index,
+                            message: _.get(assertion, 'error.message', assertion.name || ''),
+
+                            stack: 'AssertionFailure: ' + _.get(assertion, 'error.message', '') + '\n' +
```
shamasis: @kamalaknn  We are not setting this to `AssertionError` as discussed?

kamalaknn: @shamasis the `message` is set to `AssertionError` from `runtime`. Only the `stack` is constructed here. Let me update the `stack` to match the `message`.

kamalaknn:
Superceded by #1292

PR #1277, opened 2017-11-08
Support option for global timeout

harryi3t:


lib/cli/index.js
```diff
@@ -96,6 +96,8 @@ var _ = require('lodash'),
             help: 'DEPRECATED: Run a single folder from a collection. To be used with -c or -u'
         });
 
+        // @todo This is conflicting with '-r: --reporters'. Should we remove this?
+        // Also for this we have '--timeout-request'
         parser.addArgument(['-r', '--requestTimeout'], {
```
kunagpal: This set of arguments is only handled when `run` is not specified, the diff can be removed.

test/unit/cli.test.js
```diff
@@ -54,6 +54,8 @@ describe('cli parser', function () {
             '-g myGlobals.json ' +
             '-G exported_glob.json ' +
             '--delay 12000 ' +
+            // @todo We have two conflicting commands for -r (--reporters and --requestTimeout).
```
kunagpal: Remove.

PR #1278, opened 2017-11-08
Migrated from argparse to commander

deepakpathania:
- Moved from argparse to commander for CLI argument parsing.
- Restored legacy options and made them compatible with commander.
- Moved the entire CLI parsing logic to `bin/newman.js` and deleted `lib/cli/index.js`
- Added and modified tests for commander specific functionality.


bin/newman.js
```diff
@@ -2,9 +2,464 @@
...
+            throw new Error('The value must be a positive integer.');
+        }
+
+        return num.valueOf();
+    },
+
+    /**
+     * Separates reporter specific arguments from the rest.
+     *
+     * @todo Figure out a better way! (Or submit a PR to argparse).
```
kunagpal: The `argparse` clause can now be removed. 😅

bin/newman.js
```diff
@@ -2,9 +2,464 @@
...
+        memo.push(arg);
+        return memo;
+    },
+
+    /**
+     *  used for resetting the program instance for consecutive runs.
+     *
+     * @param {Object} program - The command commander program instance.
+    */
+    resetProgram = function (program) {
```
kunagpal: This function doesn't need to take in `program` as an argument.

bin/newman.js
```diff
@@ -2,9 +2,464 @@
...
+            optionNames.push(optionName);
+        });
+
+        for (prop in options) {
+            if (_.includes(optionNames, String(prop))) {
+                run[prop] = options[prop];
+            }
+        }
+
+        if (!run.collection && (options.rawArgs || options.args)) {
```
kunagpal: The if blocks here are repetitive, iterate over an array of constants instead.

bin/newman.js
```diff
@@ -276,57 +274,28 @@ var _ = require('lodash'),
...
-        if (run.html && !run.reporterHtmlExport) {
-            run.reporterHtmlExport = run.html;
-        }
-        if (run.testReportFile && !run.reporterJunitExport) {
-            run.reporterJunitExport = run.testReportFile;
-        }
-        if (run.outputFile && !run.reporterJsonExport) {
-            run.reporterJsonExport = run.outputFile;
-        }
+        args = [{ alt: 'url', target: 'collection' }, { alt: 'environmentUrl', target: 'environment' },
```
kunagpal: Since there are a lot of elements, let each be on its own line 😓 
```js
args = [
    {},
    {}
    ...
];
```

PR #1287, opened 2017-11-19
Added man page, specified in package.json

deepakpathania:
- Added a man page for newman, `man newman` :100: .


shamasis:
Umm. Which branch has the man file? Also, is this the archived variant of man page? Can you tell me how’s this .gz generated and why’s it .gz?

deepakpathania:
@shamasis Sir the `feature/man-page` branch has the man file in the `man` folder. 
I have used[ this](https://github.com/sunaku/md2man) tool to convert the markdown to manual page format(with additional formatting) and then gzipped it because man pages can transparently process gzip'ed or bzip2'ed pages, a feature we can use to free some disk space while keeping our documentation available.

npm/prepublishOnly.js
```diff
@@ -0,0 +1,26 @@
...
+// ---------------------------------------------------------------------------------------------------------------------
+// This script is intended to create man page for this module
+// ---------------------------------------------------------------------------------------------------------------------
+
+process.chdir(__dirname);
+
+var fs = require('fs'),
+    readmeToManPage = require('readme-to-man-page'),
+    path = require('path'),
+    readme = fs.readFileSync(path.join(process.cwd(), '../README.md'), 'utf8'),
```
kunagpal: Replace `process.cwd()` with `__dirname`

npm/prepublishOnly.js
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env node
+// ---------------------------------------------------------------------------------------------------------------------
+// This script is intended to create man page for this module
+// ---------------------------------------------------------------------------------------------------------------------
+
+process.chdir(__dirname);
+
+var fs = require('fs'),
+    readmeToManPage = require('readme-to-man-page'),
```
kunagpal: Group all inbuilt modules first, followed by the installed libraries.

npm/prepublishOnly.js
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env node
```
kunagpal: Make this script require friendly like all other npm scripts.

npm/prepublishOnly.js
```diff
@@ -0,0 +1,43 @@
...
+    manPageContent = readmeToManPage(readme, {
+        name: pkg.name,
+        version: pkg.version,
+        description: pkg.description,
+        section: '1',
+        manual: 'Newman User Manual'
+    });
+
+    // write the generated content to the specified path
+    fs.writeFile(manPath, manPageContent, function (err) {
```
kunagpal: This will fail if the `man` subdirectory does not already exist, add a sanity check before this.

kunagpal:
@deepakpathania Just checked out the `man` layout, the overflow is quite prevalent. This is particularly true for the API reference and reporter options table. 😓

deepakpathania:
@kunagpal I have added the sanity check.
Also, yeah the overflow, as I mentioned looks very odd due to the table layout and since the current implementation takes the content directly from the README, it cannot be altered. :disappointed: 

I have added an [issue](https://github.com/eush77/readme-to-man-page/issues/2) to the dependency repo regarding this and by the time we hear from them, we can probably come up with a workaround. Modifying README for compliance could be an option to shorten/remove the table layouts but this should be considered as a last resort.

kunagpal:
There are issues with text overflowing in tables, will have to revisit at a later date.

shamasis:
@kunagpal - Can we review this branch from your side in latest context a bit more? Especially, I removed the prebublishOnly ... maybe re-add it and see if even prepublishOnly is the right place.

shamasis:
Also - I don't see the overflow fix has been looked into by the creator. Personally, I don't see how the tool developer would do this. Readme might need change.

PS: closing this on second thought.

PR #1290, opened 2017-11-20
Update newman run documentation for URL based runs

mayank13:
newman run doesn't work with zsh without quoting the URL. Changing the docs so it works with zsh as well bash.
Changing
 newman run https://api.getpostman.com/collections/$uid?apikey=$apiKey \
    --environment https://api.getpostman.com/environments/$uid?apikey=$apiKey  

to 

 newman run "https://api.getpostman.com/collections/$uid?apikey=$apiKey" \
    --environment "https://api.getpostman.com/environments/$uid?apikey=$apiKey"


kunagpal:
L.G.T.M, pending green builds. Thanks @mayank13!

PR #1302, opened 2017-11-24
Support for missing options for runtime v7.10[WIP]

deepakpathania:
- Added support for the following options.
   * --delay-iteration 
    * --abort-on-error 
    * --block-hosts

### DIFF
--- a/bin/newman.js
+++ b/bin/newman.js
@@ -72,6 +72,22 @@ var _ = require('lodash'),
     },
 
     /**
+     *  used for converting host names to the required object format.
+     *
+     * --block-hosts 192.168.1.1
+     *
+     * @param {Array} hosts - The array of host name to block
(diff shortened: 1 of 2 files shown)


shamasis:
The parameter naming has to be decided keeping long term aspects in mind. Why do we need abort on error? We have --bail 

What about script timeout options?

shamasis:
Also, why are we deviating parameter names so much from the runtime option names?

deepakpathania:
@shamasis Sir, the option names were chosen such that they have a more user-friendly and self-descriptive name as a flag whereas a more technical one for the runtime, as described by @kunagpal.
 
I added the WIP since not all options to be added were described completely yet and thought would keep on adding them by pushing to this branch. Sorry for the confusion.

PR #1310, opened 2017-11-29
Added --abort-on-error option

deepakpathania:


bin/newman.js
```diff
@@ -202,6 +202,7 @@ var _ = require('lodash'),
                 'Specify the path to the Client SSL key (not needed for .pfx files)')
             .option('--ssl-client-passphrase <path>',
                 'Specify the Client SSL passphrase (optional, needed for passphrase protected keys).')
+            .option('--abort-on-error', 'Abruptly halts the run on errors, and directly calls the done callback')
```
kunagpal: This description needs to be made more Newman friendly.

lib/run/index.js
```diff
@@ -92,6 +92,7 @@ module.exports = function (options, callback) {
 
         // now start the run!
         runner.run(options.collection, {
+            abortOnError: options.abortOnError,
             stopOnFailure: options.bail, // LOL, you just got trolled ¯\_(ツ)_/¯
```
kunagpal: Add a fallback to `options.abortOnError` as well.

shamasis:
Option name not thought through. There is already a `--bail`. This should have been a qualifier option to the bail argument, which adjusted the behaviour of when to bail.

PR #1320, opened 2017-12-01
Support modifiers with bail option

deepakpathania:
- Adds support for adding modifiers to the bail option.
- Maps the added modifiers to runtime options. 

For eg. Something like 
`newman run collection.json --folder invalidName --bail` used to gracefully exit the collection run but would go through the intermediate steps.

Now with modifiers, we can skip the intermediate steps altogether and end the collection run with an error by directly calling the done callback. 

`newman run collection.json --folder invalidName --bail folder` ends the collection run directly with `Unable to find a folder or request: invalidName`


lib/run/index.js
```diff
@@ -92,8 +110,10 @@ module.exports = function (options, callback) {
 
         // now start the run!
         runner.run(options.collection, {
-            stopOnFailure: options.bail, // LOL, you just got trolled ¯\_(ツ)_/¯
+            stopOnFailure: stopOnFailure, // LOL, you just got trolled ¯\_(ツ)_/¯
             abortOnFailure: options.abortOnFailure, // used in integration tests, to be considered for a future release
+            abortOnError: typeof options.bail !== 'undefined' && options.bail.folder ?
```
kunagpal: _.get(options, 'bail.folder')

test/library/run-options.test.js
```diff
@@ -98,6 +98,31 @@ describe('Newman run options', function () {
         });
     });
 
+    describe('bail modifiers', function () {
+        it('should skip collection run in case of error when folder is specified', function (done) {
+            newman.run({
+                collection: 'test/fixtures/run/failed-request.json',
+                bail: ['folder']
```
kunagpal: For this test to make sense, the folder option should also be specified.

test/library/run-options.test.js
```diff
@@ -98,6 +98,31 @@ describe('Newman run options', function () {
...
+            }, function (err) {
+                expect(err).to.be.ok();
+                expect(err.message).to.be('getaddrinfo ENOTFOUND 123.random.z 123.random.z:443');
+
+                done();
+            });
+        });
+
+        it('should gracefully stop a collection run in case of error with no additional modifiers', function (done) {
```
kunagpal: Also add a test where both `folder` and `failure` are present as `bail` options.

lib/run/index.js
```diff
@@ -92,8 +110,10 @@ module.exports = function (options, callback) {
 
         // now start the run!
         runner.run(options.collection, {
-            stopOnFailure: options.bail, // LOL, you just got trolled ¯\_(ツ)_/¯
+            stopOnFailure: stopOnFailure, // LOL, you just got trolled ¯\_(ツ)_/¯
```
kunagpal: Can be simplified as:
```
stopOnFailure: _.get(options, 'bail.failure', options.bail)
```

lib/run/index.js
```diff
@@ -84,6 +85,23 @@ module.exports = function (options, callback) {
...
+            options.bail.forEach(function (option) {
+                obj[option] = true;
+            });
+            options.bail = obj;
+        }
+
+        // sets stopOnFailure to true in case bail is used without any modifiers or with failure
+        // --bail => stopOnFailure = true
+        // --bail failure => stopOnFailure = true
+        (typeof options.bail !== 'undefined' &&
```
kunagpal: See comment below.

lib/run/index.js
```diff
@@ -84,6 +85,23 @@ module.exports = function (options, callback) {
             return callback(new Error('newman: expecting a collection to run'));
         }
 
+        // iterates over the bail array and sets each item as an obj key with a value of boolean true
+        // [item1, item2] => {item1: true, item2: true}
+        if (_.isArray(options.bail)) {
```
kunagpal:
```js
if (_.isArray(options.bail)) {
    options.bail = _.transform(options.bail, function (result, value) {
        result[value] = true;
    }, {});
}
```

bin/newman.js
```diff
@@ -178,8 +190,10 @@ var _ = require('lodash'),
             .option('--export-globals <path>', 'Specify an output file to dump Globals before exiting.')
             .option('--export-collection <path>', 'Specify an output file to save the executed collection')
             .option('--delay-request [n]', 'Specify the extent of delay between requests (milliseconds)', Integer, 0)
-            .option('--bail',
-                'Specify whether or not to gracefully stop a collection run on encountering the first error.')
+            .option('--bail [modifiers]',
```
kunagpal: `[optional modifiers]`

bin/newman.js
```diff
@@ -71,6 +71,18 @@ var _ = require('lodash'),
...
+    /**
+     *  used for converting a comma separated string to an array.
+     *
+     * eg. item1,item2
+     *
+     * @param {String} list - The comma separated string.
+     * @returns {Array} - [item1, item2] - The array representation of the passed string.
+    */
+    arrayCollect = function (list) {
+        return list.split(',');
```
kunagpal:
```js
return _.split(list, ',');
```

bin/newman.js
deepakpathania: `[ ]` mean optional only, `<>` these mean required.

lib/run/index.js
deepakpathania: Doesn't work, sets `stopOnFailure` to `{ folder: true }` in case of `--bail folder`

test/library/run-options.test.js
deepakpathania: This basically just checks for error and whether or not the bail option affects it, specifying folder option will not change the core test behavior.

test/library/run-options.test.js
```diff
@@ -98,6 +98,43 @@ describe('Newman run options', function () {
...
+                expect(err).to.not.be.ok();
+
+                done();
+            });
+        });
+
+        it('should skip collection run in case of error when both folder and failure are specified', function (done) {
+            newman.run({
+                collection: 'test/fixtures/run/failed-request.json',
+                bail: ['folder', 'failure']
```
kunagpal: Specify the `folder` option here to ensure that the invalid folder check is actually working 😓

PR #1350, opened 2017-12-19
Adds run flow change information to CLI reporter

deepakpathania:
- Intimates user about run flow change information caused by `postman.setNextRequest`.
- This is how it looks in action. 

selection_030


lib/reporters/cli/index.js
```diff
@@ -13,7 +13,8 @@ var _ = require('lodash'),
     DOT = '.',
     E = '',
 
-    PostmanCLIReporter;
```
kunagpal: In the future, let's place all such variable additions above, so the diff is cleaner 😄

lib/reporters/cli/index.js
```diff
@@ -24,6 +25,19 @@ colors.setTheme({
     error: 'red'
 });
 
+extractSNR = function (executions) {
+    var snr = {};
+
+    _.isArray(executions) && executions.forEach(function (execution) {
```
kunagpal: If the intent is to find the last existing definition of `nextRequest`, use `_.forEachRight` instead, with a bailout on the first hit.

lib/reporters/cli/index.js
```diff
@@ -24,6 +25,19 @@ colors.setTheme({
...
+        _.has(_.get(execution, 'result.return'), 'nextRequest') && (
+            (snr.defined = true),
+            (snr.value = execution.result.return.nextRequest)
+        );
+    });
+
+    return snr;
```
kunagpal: It's cleaner to return the matched value directly. Use that in the condition below as well.

lib/reporters/cli/index.js
```diff
@@ -89,6 +103,16 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
         print.lf(LF + colors.gray.underline('Iteration %d/%d'), o.cursor.iteration + 1, o.cursor.cycles);
     });
 
+    emitter.on('test', function (err, o) {
+        if (err) {
+            return;
+        }
+        var snr = extractSNR(o.executions);
+        if (_.get(snr, 'defined')) {
```
kunagpal: See comment above.

lib/reporters/cli/index.js
```diff
@@ -89,6 +103,16 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
...
+            print.lf(LF + colors.gray('Attempting to set next request to', snr.value));
```
kunagpal: @shamasis Is the copy fine?

kunagpal:
@deepakpathania The changes are fine, I'd still like @shamasis to take a look at the copy.

Issue #1334, opened 2017-12-11
Assertion error description


BBleyen:
The AssertionError description changed to an other output
the test: tests[\"response code is 200\"] = responseCode.code === 200
Now: Version 3.9.0
 #  failure         detail      
 1.  AssertionError  expected false to be truthy   
                     at assertion:0 in test-script 
                     inside ""

Previous
 #  failure        detail
 1\.  AssertionFai…  response code is 200
                    at assertion:1 in test-script
                    inside ""

Is there a command to trigger the previous output or an option to add that information to the detailed failures info.
So far I have not found any info regarding this change or behaviour.

Kind regards

kunagpal:
@BBleyen Thanks for reporting this. While a fix is worked on, could you use the new `pm.*` equivalent instead?
```
// old
tests['response code is 200'] = responseCode.code === 200;

// new
pm.test('response code is 200', function () {
  pm.response.to.be.ok;
});
```

BBleyen:
@kunagpal thank you for your fast reply and alternative solution
It is not that urgent, so I will not change all of my tests to the other format.
But I will use the new format for the new ones that I am working with.

efrenette11:
This is for me a bug that cause the issue investigationg hard. I'm using newman, and from the repport. I'm not able to figure out what was the problem.

When should we expect to have a fix. I can't use pm.test. 

here a code test.

var a = "a"
var b = "b"


tests["test A : "] = a === b;





![image](https://user-images.githubusercontent.com/30326999/34832207-3a4ef634-f6b7-11e7-930c-9d5b0ffdbc45.png)

efrenette11:
Also have this issue in Version of postman

efrenette11:
Seem to be the same https://github.com/postmanlabs/postman-app-support/issues/3888. But was closed under canary version 03 ???

kunagpal:
@BBleyen @efrenette11 Apologies for the delay, a fix is being worked on in #1390, and should be available with the next Newman release. 😄

PR #1390, opened 2018-01-19
Fixed legacy test name reporting

kunagpal:
With this change, actual legacy test names are shown in CLI failure reports, instead of just `expected false to be truthy`.

Takes care of https://github.com/postmanlabs/newman/issues/1334


lib/run/index.js
```diff
@@ -210,6 +210,7 @@ module.exports = function (options, callback) {
                         emitter.emit('assertion', (assertion.passed ? null : {
                             name: errorName,
                             index: assertion.index,
+                            test: assertion.name,
```
shamasis: This needs to be added as test. Isn’t it?

kunagpal: Added

PR #1428, opened 2018-02-14
Fixed help text issue

deepakpathania:


shamasis:
@deepakpathania can you elaborate on the reason of this change in the PR description?

deepakpathania:
@shamasis Sir currently, when no arguments are supplied to newman, it doesn't show help text and rather shows the following message.
```
newman: the v2.x CLI options are deprecated. You should use newman run <path> [options] instead.
        refer https://github.com/postmanlabs/newman/blob/develop/MIGRATION.md for details.
the file at /usr/local/bin/newman does not contain valid JSON data
Unexpected token '#' at 1:1
#!/usr/bin/env node
^
```
This happens because the check for `legacyMode` doesn't account for this condition and thus we consider the situation with no supplied arguments as `legacyMode`. We have a check for showing the help text with no supplied arguments but that accounts for only non-legacy mode cases. This adds a check to handle this case.

bin/newman.js
```diff
@@ -401,7 +401,8 @@ var _ = require('lodash'),
      */
     rawOptions = function (procArgv, programName, callback) {
         var legacyMode = !_.includes(procArgv, 'run') &&
-        !_.includes(['--help', '-h', '--version', '-v', '-V'], procArgv[2]),
+        !_.includes(['--help', '-h', '--version', '-v', '-V'], procArgv[2]) &&
+        !_.isEmpty(procArgv.slice(2)),
```
kunagpal: `procArgv.slice(2)` is being used on line 415 below as well, reference its value from a variable.

PR #1450, opened 2018-02-28
feat: Add noSuccessAssertions

stephan-nordnes-eriksen:
It would be very nice to be able to ignore successful assertions.

I have a very large test-suite, and my errors disappears in a wall of successes which makes it hard(er) to debug. This commit/PR adds a simple and naive `noSuccessAssertions` option to the cli reporter.


shamasis:
@stephan-nordnes-eriksen - great idea. the tests failed - guess network related reasons. I have restarted the build.

The PR has to be completed by:

1. We need to have CLI test to ensure we don't accidentally break this later

2 The if-block could be moved up top with an early bailout on fail or skip in case this option is turned on. Will save a check on every assertion and would have a cleaner exit from function.

3. We need to add the change to the README.

stephan-nordnes-eriksen:
Sorry for the slow response. Have been traveling. I will get going on this as soon as possible

kunagpal:
Hey @stephan-nordnes-eriksen, could you address the review comments? 😄

stephan-nordnes-eriksen:
Hey @kunagpal 

Sorry for taking forever. I am looking at these now. I am looking at the other cli tests to see what I can base my tests on, but it appears to me that there are no explicit tests for any of the `reporter-cli-`-parameters. Perhaps I don't know what to look for? Some quick guidance would be much appreciated.

I can perhaps add `--reporter-cli-no-success-assertions` together with `--reporter-cli-no-summary` in `cli.test.js`, but I am not sure it will help much.

Anyways. I have made some updates to the readme and stuff. Will upload soon.

kunagpal:
No worries, @stephan-nordnes-eriksen. A good test for `no-success-assertions` would be to assert that the generated CLI report does not contain any text from the successful assertions, like so:

Working examples can be found in https://github.com/postmanlabs/newman/blob/develop/test/cli/color-tty.test.js, where tests have been added for the validity of CLI colour options. You'll have to add the tests here: https://github.com/postmanlabs/newman/blob/develop/test/cli/cli-reporter-failure-details.test.js

test/unit/cli.test.js
```diff
@@ -227,6 +227,7 @@ describe('cli parser', function () {
             '--export-environment exported_env.json ' +
             '--export-globals exported_glob.json ' +
             '--reporter-cli-no-summary ' +
+            '--reporter-cli-no-success-assertions ' +
```
kunagpal: This will also have to be tested (see the sequence of `expect` statements below) 😄

test/cli/cli-reporter-no-success-assertsions.test.js
```diff
@@ -0,0 +1,24 @@
+describe('CLI reporter no success assertions', function () {
+    var noSuccessOutput = /✓\s+response\s+code\s+is\s+200/;
```
kunagpal: The `✓` at the beginning of this pattern is causing Windows tests to fail :/. You'll have to remove it.

stephan-nordnes-eriksen: Ahh. Yeah, will remove it. Still working on the cli.test.js.

test/unit/cli.test.js
stephan-nordnes-eriksen: This should be done now. Can you check it?

PR #1495, opened 2018-04-11
Adding feature to disable "snake-looking" output when waiting for response

pontusarfwedson:
When running newman in a C/I system we get console output presented in our browser. Some browser to not support seem to support the `BS = '\b'` which is used to continuously remove the last character (and create the flowing feel when waiting). Hence my PR adds the run option `--disable-wait-output` which will suppress the "snake"-output when waiting for http response and instead show nothing. This is intended for use cases where people are not staring at the terminal "live" during the run but rather are looking at the output in retrospect.

### DIFF
--- a/bin/newman.js
+++ b/bin/newman.js
@@ -159,6 +159,7 @@ var _ = require('lodash'),
             .option('--silent', 'Prevents newman from showing output to CLI', false)
             .option('--color', 'Force colored output (for use in CI environments)')
             .option('collection <path>', 'URL or path to a Postman Collection')
+            .option('--disable-wait-output', 'Remove the wait output that uses backslash b to reduce unnecessary output in some environments')
             .parse(rawArgs);
     },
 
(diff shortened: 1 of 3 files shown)


kunagpal:
@pontusarfwedson Did you try running Newman with the `--disable-unicode` option?

pontusarfwedson:
Yes, it still gives me `⠄⠆⠇⠋⠙⠸⠰⠠⠰⠸⠙⠋⠇⠆⠄⠆⠇⠋⠙⠸⠰⠠⠰⠸⠙⠋⠇⠆⠄⠆⠇⠋⠙⠸⠰⠠⠰⠸⠙⠋⠇⠆` and someone on unix with chrome gets each char on new line. So no improvement with that option.

kunagpal:
@pontusarfwedson The waiting snake spinner is added here: https://github.com/postmanlabs/newman/blob/develop/lib/reporters/cli/index.js#L137, which can be changed to incorporate the `--disable-unicode` option as well.

pontusarfwedson:
Alright, but is this really a unicode thing?

pontusarfwedson:
Added another pull request which makes that simple change so the spinner is removed from the emitter when have `--disable-unicode`. Not sure that unicode is exactly what it is but if you prefer that then I can close this PR and we can look at the other one as well :)

kunagpal:
Thanks, @pontusarfwedson! I've left comments on the new pull request. Closing this in favour of #1496

PR #1496, opened 2018-04-11
added so --disable-unicode removes spinner when waiting for http resp…

pontusarfwedson:


lib/reporters/cli/index.js
```diff
@@ -134,7 +134,12 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
     emitter.on('beforeRequest', function (err, o) {
         if (err) { return; }
```
kunagpal: This can be cleanly written as:
```js
        if (err || !o.request) { return; }

        print('  %s %s ', colors.gray(o.request.method), colors.gray(o.request.url));

        !options.disableUnicode && print().wait(colors.gray);

```

pontusarfwedson:
Changed to conform to your nit-pick comment :)

Issue #1527, opened 2018-05-10
Default HTML report template multiplies folder and request IDs


bartmanus:
## Problem description

Default HTML report template at [newman/lib/reporters/html/template-default.hbs](https://github.com/postmanlabs/newman/commits/develop/lib/reporters/html/template-default.hbs) reuses folder and request IDs between iterations:

* Reuses folder ID (`parent.id`) for collapsible folderData `div` in the `{{#each aggregation}}` block
  * causing report with multiple iterations to have the same folder ID in each iteration
  * causing a click on any folderHead to always collapse/uncollapse the first occurrence (in the first iteration) of the folderData in the whole report.
* reuses request ID (`item.id`) for collapsible requestData `div` in the `{{#each executions}}` block
  * causing report with multiple iterations to have the same request ID in each iteration
  * causing a click on requestHead to always collapse/uncollapse the first occurrence (in the first iteration) of the requestData in the whole report.

## Environment

 1. Newman Version: 3.9.3
 2. OS details (type, version, and architecture): Centos 7 x86_64 container host, newman Docker image ([postman/newman_alpine33](https://hub.docker.com/r/postman/newman_alpine33/):latest ID=sha256:3ced17a46f83d133f16c613414b0fbd0a713fb9ea8a39f61824611f238bd842e)
 3. Are you using Newman as a library, or via the CLI? CLI
 3. Did you encounter this recently, or has this bug always been there: noticed yesterday
 4. Expected behaviour: the folder or request from the current iteration should collapse
 5. Command / script used to run Newman: `newman run postman_collection.json --iteration-count 3 --suppress-exit-code --reporters html,junit`
 6. Sample collection, and auxiliary files (minus the sensitive details):

## Proposed solution

Append iteration index from the `{{#each aggregations}}` block into mentioned IDs:

* `-agg{{@index}}` for folderHead/folderData `div` s
* `-agg{{@../index}}` for requestHead/requestData `div` s

I will create a fork with these fixes jand submit a pull request.

kunagpal:
@bartmanus This has been fixed with Newman v3.9.4 😄

kunagpal:
@bartmanus A quick update: This is by design, as the HTML reporter is meant to display aggregated request results. The original P.R was merged in haste, I've issued a revert in #1552

kunagpal:
@bartmanus This has been fixed in Newman v3.10.0 😄

PR #1553, opened 2018-05-26
Corrected request aggregation logic in the HTML reporter

kunagpal:
* A (potentially long-standing) bug prevented the HTML reporter from aggregating results correctly when collections with folders were run with multiple iterations. This has now been fixed.

Discovered in #1527

### DIFF
--- a/lib/reporters/html/index.js
+++ b/lib/reporters/html/index.js
@@ -58,6 +58,7 @@ PostmanHTMLReporter = function (newman, options) {
             netTestCounts = {},
             aggregations = [],
             traversedRequests = {},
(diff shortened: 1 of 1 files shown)


Kal6:
Hi @kunagpal , when can we expect this fix? i spent several days why it was happening but now i realized there is a bug after going through all of these threads  (lol). 
Thanks you - Kal

kunagpal:
Hey @Kal6, thanks for reaching out. There are a few pending bugfixes, which will be merged over the course of the next week. You can expect a release shortly after that. 😄

Issue #1547, opened 2018-05-25
list failures stack in the html report


enesyalcin:
It would be nice to have the failures stack in the html report.

![image](https://user-images.githubusercontent.com/17004938/40541837-cb220352-601d-11e8-9da9-661e87d9e38a.png)

This is not enough to grasp the exact error. Therefore the Information which is at the Bottom of the cli Output is also needed in the html report.

![image](https://user-images.githubusercontent.com/17004938/40541922-19534fd6-601e-11e8-9bc2-7a4f8755c8d5.png)

If I am not mistaken, currently only the length of the failures is being transmitted to the html report. It would Need to be Extended by an additional property with the failures object in it.

https://github.com/postmanlabs/newman/blob/136993d180953fe98222211255319ab639ad4c49/lib/reporters/html/index.js#L164

MarcosEllys:
Hey @enesyalcin I had the same problem when I created my custom template and I shared it here https://github.com/MarcosEllys/newman-html-report-template-verbose.

If you want, you can test it

kunagpal:
@enesyalcin This has been fixed in Newman v3.10.0 😄

PR #1555, opened 2018-05-26
HTML reporter failures list

kunagpal:
#1547 

screen shot 2018-05-26 at 10 02 38 pm


kunagpal:
@shamasis Good spot, but handlebars will safely handle all such property accesses. There's even a test that checks for proper HTML report generation with errored collection runs.

Issue #1562, opened 2018-06-01
newman globals exported with --export-globals cannot be loaded into postman


connorsadlervelo:
newman: 3.9.4
postman: Version 6.1.3 (6.1.3)
Macbook Pro 16 GB , Sierra 10.12.6

I am running newman with "--export-globals" to produce a .json file containing global variable values.
I then try to Import that file into Postman but get an error: 
Error while importing: format not recognized.

After some digging (comparing with a globals file exported from Postman) I found the "name" field is missing in the globals file created by newman.
If I add in the following line near the bottom of the file created by newman, it imports into Postman OK:
   "name": "XXX",

Should the file formats be identical?

Regards,

Connor

**Script to run newman:**
newman run "CFS export globals test.postman_collection.json" --bail failure -r cli --export-globals "CFS export globals test - exported globals from newman.json"

**Collection: CFS export globals test.postman_collection.json**
```
{
	"info": {
		"_postman_id": "38c0e9de-af67-4e1a-9e43-fec47fa5b868",
		"name": "CFS export globals test",
		"schema": "https://schema.getpostman.com/json/collection/v2.1.0/collection.json"
	},
	"item": [
		{
			"name": "Change Global 1",
			"event": [
				{
					"listen": "test",
					"script": {
						"id": "a666463b-75d4-4d0b-ab9a-c0f766516d54",
						"type": "text/javascript",
						"exec": [
							"postman.setGlobalVariable(\"myGlobal1\", \"hello\")",
							"postman.setGlobalVariable(\"myGlobal2\", \"world\")"
						]
					}
				}
			],
			"request": {
				"method": "GET",
				"header": [],
				"body": {},
				"url": {
					"raw": "https://postman-echo.com/delay/0",
					"protocol": "https",
					"host": [
						"postman-echo",
						"com"
					],
					"path": [
						"delay",
						"0"
					]
				}
			},
			"response": []
		}
	]
}

```

connorsadlervelo:
For completeness:

**File exported from newman:**
```
{
  "id": "d51618d9-0c34-407e-8d3f-bba720dddaa2",
  "values": [
    {
      "type": "any",
      "value": "hello",
      "key": "myGlobal1"
    },
    {
      "type": "any",
      "value": "world",
      "key": "myGlobal2"
    }
  ],
  "_postman_variable_scope": "globals",
  "_postman_exported_at": "2018-06-01T16:36:21.593Z",
[...]

PR #1563, opened 2018-06-01
Added fallback names in variable file exports

kunagpal:
Discovered in https://github.com/postmanlabs/newman/issues/1562

In cases where the collection is run without starter environment/global variable files, but contains variable setter statements, exported environment/global files will not have the name property. This causes app-import validations to fail, thereby preventing imports.


shamasis:
Can you briefly describe the intended behaviour in the PR so that I can cross check with the code and tests?

kunagpal:
@shamasis Added a description above. 😄

PR #1567, opened 2018-06-04
junit reporter not working

saiharish:
Junit reporter does not work in the version 3.9.4. When run with the command **_--reporters junit_** the xml file generated always contains tests 0. The tests are run but the junit report is not generated. json and html report work fine.


kunagpal:
@saiharish I'm unable to reproduce this with Newman v3.9.4, and the sample collection here: https://github.com/postmanlabs/newman/blob/develop/examples/sample-collection.json.

Also, it looks like you wanted to create an issue, but ended up creating a pull request instead 😄

saiharish:
@kunagpal Thanks for the reply. By mistake I created pull request instead of issue. Also, the junits report worked. I have not created any tests in my collection.json.

PR #1570, opened 2018-06-06
Use node 10

rawaludin:
Using node 8, I'm getting this error on test script:

```
Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
```


kunagpal:
@rawaludin This is because the v3.9.4 release for the Alpine Docker image uses Node v4. The next release will address this issue, as Node v8 has been added to the Alpine image.

rawaludin:
Got it. Is there some place I could subscribe for getting notification for next release?
Thanks!

PR #1576, opened 2018-06-09
Corrected pre-existing directory handling for exports

### LINKED ISSUE #643: reporter-junit-export does not handle directories
The documentation for the reporter-junit-export command line argument states:

> Specify a path where the output XML file will be written to disk.

When setting it to a directory with `--reporter-junit-export /etc/newman/foo`, newman reports:

> error writing file "/etc/newman/foo" for junit-reporter
> EISDIR: illegal operation on a directory, open '/etc/newman/foo'

If the directory does not exists, a file called /etc/newman/foo with the junit report gets created.

IMHO newman should handle the argument like:
- if it's a directory, create the files with the default naming scheme in this directory
- if it's a file, create the file with this name

kunagpal:
Fixes #643 

If the target export path is a pre-existing directory, the target file will be created inside it, instead of throwing an `EISDIR` error.


lib/run/export-file.js
```diff
@@ -77,9 +82,25 @@ module.exports = function (options, done) {
                     `error creating path for file "${path.unparsed}" for ${options.name || 'unknown-source'}`));
             }
 
-            fs.writeFile(path.unparsed, content, function (err) {
-                done(_.set(err, 'help',
-                    `error writing file "${path.unparsed}" for ${options.name || 'unknown-source'}`), path);
+            fs.stat(path.unparsed, function (error, stat) {
+                var target;
+
```
shamasis: Error handling? Not sure what to do here. Need to decide. Suggestions?

lib/run/export-file.js
```diff
@@ -77,9 +82,25 @@ module.exports = function (options, done) {
...
+                // handle cases where the specified export path is a pre-existing directory
+                if (stat && stat.isDirectory()) {
```
shamasis: If stat is a file?

lib/run/export-file.js
```diff
@@ -77,9 +82,25 @@ module.exports = function (options, done) {
                     `error creating path for file "${path.unparsed}" for ${options.name || 'unknown-source'}`));
             }
 
-            fs.writeFile(path.unparsed, content, function (err) {
-                done(_.set(err, 'help',
-                    `error writing file "${path.unparsed}" for ${options.name || 'unknown-source'}`), path);
+            fs.stat(path.unparsed, function (error, stat) {
```
shamasis: Should we check these on start?

lib/run/export-file.js
kunagpal: Then it gets overwritten, which has been the behaviour since exports were introduced.

lib/run/export-file.js
kunagpal: As in, before the `mkirp` call, right?

lib/run/export-file.js
kunagpal: If there's an error (the path doesn't exist), then it gets written to as usual. This could be a place to add debug logs, though.

lib/run/export-file.js
kunagpal: @shamasis I tried that, it ends up looking a bit repetitive:
```
        fs.stat(path.unparsed, function (error, stat) {
            if (error) {
                return mkdirp(path.dir, function (err) {
                    if (err) {
                        return done(_.set(err, 'help',
                            `error creating path for file "${path.unparsed}" for ${options.name || 'unknown-source'}`));
                    }

                    fs.writeFile(path.unparsed, content, function (err) {
                        done(_.set(err, 'help',
                            `error writing file "${path.unparsed}" for ${options.name || 'unknown-source'}`), path);
                    });
                });
            }

            var target;

            // handle cases where the specified export path is a pre-existing directory
            if (stat && stat.isDirectory()) {
                target = parsePath(options.default);

                // append timestamp
                // @todo make -0 become incremental if file name exists
                target.name += '-' + timestamp() + '0';
                target.base = target.name + target.ext;

                path.unparsed = joinPath(path.unparsed, target.base);
            }

            fs.writeFile(path.unparsed, content, function (err) {
                done(_.set(err, 'help',
                    `error writing file "${path.unparsed}" for ${options.name || 'unknown-source'}`), path);
            });
        });
```

PR #1585, opened 2018-06-15
CLI reporter terminal width

### LINKED ISSUE #1203: feature request to turn off text wrapping or force a width larger than 60 for console.log events

kunagpal:
Resolves #1203


shamasis:
I guess to full cli would be crazy! But maybe it’s okay for CLI tables. By that, it would logically make sense that it’s applied to formatted elements and not continuous lines. Thoughts?

shamasis:
@kunagpal updates?

PR #1589, opened 2018-06-19
Migrate Library Tests to Chai

codenirvana:
- Use chai expect assertions
- Move chai to devDependencies


package.json
```diff
@@ -64,6 +63,7 @@
     "xmlbuilder": "10.0.0"
   },
   "devDependencies": {
+    "chai": "4.1.2",
```
kunagpal: 👏

PR #1594, opened 2018-06-20
CLI: Properly parse --reporter- argument

### LINKED ISSUE #1590: --reporter-junit-export syntax is inconsistent (and not documented?)
Other command line options like `--environment` accepts an `=` delimiter, whereas `--reporter-junit-export` does not. Why is this? Where is this documented? Only after checking out the repository and reading MIGRATION.md was this apparent.

For example, in the command line `newman run "test/postman/newman/v1.postman_collection.json" --environment="test/postman/newman/environments/Quality.postman_environment.json" --reporters="junit,cli" --reporter-junit-export=target/test-reports/TEST-newman-report.xml` the environment is correctly picked up, reporters both are applied, but the reporter-junit-export goes to the default path and filename instead of the specified path and filename.

TIA

codenirvana:
Fixes #1590

### DIFF
--- a/bin/newman.js
+++ b/bin/newman.js
@@ -31,18 +31,33 @@ var _ = require('lodash'),
     separateReporterArgs = function (allArgs) {
         var reporterArgs = [],
             args = [],
+            arg,
+            index,
             i;
 
         // Separate the reporter arguments from the rest
         for (i = 0; i < allArgs.length; i++) {
-            if (_.startsWith(allArgs[i], '--reporter-')) {
-                reporterArgs.push(allArgs[i]);
+            // Remove trailing whitespace
+            arg = _.trim(allArgs[i]);
 
+            if (!_.startsWith(arg, '--reporter-')) {
+                args.push(arg);
+                continue;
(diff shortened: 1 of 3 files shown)


bin/newman.js
```diff
@@ -31,18 +31,33 @@ var _ = require('lodash'),
...
+
+            // Remove "--reporter-"
+            arg = arg.replace('--reporter-', '');
 
+            if (allArgs[i + 1] && !_.startsWith(allArgs[i + 1], '-')) {
                 // Also push the next parameter if it's not an option.
-                allArgs[i + 1] && !_.startsWith(allArgs[i + 1], '-') && reporterArgs.push(allArgs[++i]);
+                reporterArgs.push(arg, allArgs[++i]);
+            }
+            else if (~(index = arg.indexOf('='))) {
```
kunagpal: Handle this first, so as to avoid touching existing code to all extents possible.

Also,
```js
index = arg.indexOf('=');
if (index !== -1) {
   ...
}
```

PR #1601, opened 2018-06-29
Update reporter load strategy

codenirvana:
- Load external reporter before local reporter
- External reporter can overrides local reporter with same name

### DIFF
--- a/lib/run/index.js
+++ b/lib/run/index.js
@@ -45,7 +45,14 @@ var fs = require('fs'),
         junit: require('../reporters/junit'),
         progress: require('../reporters/progress'),
         emojitrain: require('../reporters/emojitrain')
-    };
+    },
+
+    /**
+     * The list of valid custom reporters
+     *
+     * @type {Array}
+     */
+    validCustomReporters = ['teamcity'];
 
 /**
  * Runs the collection, with all the provided options, returning an EventEmitter.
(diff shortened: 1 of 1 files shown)


lib/run/index.js
```diff
@@ -45,7 +45,14 @@ var fs = require('fs'),
         junit: require('../reporters/junit'),
         progress: require('../reporters/progress'),
         emojitrain: require('../reporters/emojitrain')
-    };
+    },
+
+    /**
+     * The list of valid custom reporters
```
shamasis: Needs to be `@private`

lib/run/index.js
```diff
@@ -45,7 +45,14 @@ var fs = require('fs'),
...
+     *
+     * @type {Array}
+     */
+    validCustomReporters = ['teamcity'];
```
shamasis: rename to 

```
knownNewmanReporters = {
    teamcity: '   run `npm install teamcity`'
};
```

lib/run/index.js
```diff
@@ -336,6 +344,10 @@ module.exports = function (options, callback) {
                 if (!(_.isFunction(Reporter) && _.isObject(emitter.reporters[reporterName]))) {
                     // @todo: route this via print module to respect silent flags
                     console.warn(`newman warning: "${reporterName}" reporter could not be loaded.`);
+                    // print install instruction in case a valid custom reporter is missing
+                    if (_.includes(validCustomReporters, reporterName)) {
+                        console.warn(`  run 'npm install newman-reporter-${reporterName}' to install`);
```
shamasis: This is a very authoritative message and as such, the messaging needs to be changed. For `@namespace/newman-reporter-reportername`, this would be misleading.

Kindly store message in object and use from there and wherever preset message is not found, add a message saying "please install reporter using npm".

PR #1610, opened 2018-07-02
Drop support for v2 CLI Options

codenirvana:


sheluchin:
Is there somewhere we can read more about this deprecation?

kunagpal:
Hey @scheluchin, there won't be any breaking changes if you're using the newer `newman run ...` command. While there isn't a full fledged migration guide just yet, you can take a look at https://github.com/postmanlabs/newman/blob/develop/MIGRATION.md, which has details on the differences between legacy and contemporary arguments. :smile:

kunagpal:
Merging this, there will be a separate pull request dedicated to housekeeping.

PR #1612, opened 2018-07-02
JUnit reporter fix

kunagpal:


lib/reporters/junit/index.js
```diff
@@ -27,6 +27,7 @@ JunitReporter = function (newman, reporterOptions) {
 
         root = xml.create('testsuites', { version: '1.0', encoding: 'UTF-8' });
         root.att('name', collection.name);
+        root.att('tests', _.get(newman, 'summary.run.stats.tests.total', 0));
```
shamasis: cannot be `0` default. set to `unknown`. else an error will give false confidence

PR #1642, opened 2018-07-29
Improved Docker postpublish script to account for master branch

kunagpal:
This will prevent the latest tag from being added to the images if the branch is not `master`.


kunagpal:
@shamasis Updated.

PR #1623, opened 2018-07-11
Refactor Newman CLI

codenirvana:

PR #1657, opened 2018-08-03
Final Newman v4 cleanup

codenirvana:


bin/util.js
```diff
@@ -75,10 +75,7 @@ module.exports = {
             validProp && (result[key] = value);
 
             return result;
-        }, {
```
codenirvana: These two properties are added unnecessarily.

Reference: https://github.com/postmanlabs/newman/pull/1623/files#diff-024316d2f6b434a9ff0a176917b4ebb3L254

lib/index.js
```diff
@@ -1,4 +1,3 @@
 module.exports = {
-    run: require('./run'),
-    version: require('./version')
```
codenirvana: Not required because of: https://github.com/postmanlabs/newman/pull/1623/files#diff-024316d2f6b434a9ff0a176917b4ebb3R12

PR #1660, opened 2018-08-06
 Drop support for v1 collection format

codenirvana:

PR #1669, opened 2018-08-08
 Deprecate support for the v1 collection format

codenirvana:
Reverts #1660

### DIFF
--- a/lib/run/index.js
+++ b/lib/run/index.js
@@ -347,7 +347,7 @@ module.exports = function (options, callback) {
                 // if the reporter is not loaded then we need to warn users
                 if (!(_.isFunction(Reporter) && _.isObject(emitter.reporters[reporterName]))) {
                     // @todo: route this via print module to respect silent flags
-                    console.warn(`newman warning: "${reporterName}" reporter could not be loaded.`);
+                    console.warn(`newman: "${reporterName}" reporter could not be loaded.`);
                     // print install instruction in case a known reporter is missing
                     if (knownReporterErrorMessages[reporterName]) {
                         console.warn(knownReporterErrorMessages[reporterName]);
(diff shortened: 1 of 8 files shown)


lib/run/options.js
```diff
@@ -65,21 +74,25 @@ var _ = require('lodash'),
...
      */
-    validateCollection = function (collection, callback) {
-        var error = null;
-
+    processCollection = function (collection, callback) {
         if (util.isV1Collection(collection)) {
-            error = new Error('Newman >= v4 does not support the v1 collection format');
-            error.friendly = 'Use the Postman Native app to export collections in the v2 format';
+            // @todo: route this via print module to respect silent flags
+            console.warn('newman warning: Newman >= v4 deprecates support for the v1 collection format');
```
kunagpal: `newman:`

kunagpal: `Newman v4`

test/unit/run.test.js
```diff
@@ -31,36 +31,6 @@ describe('run module', function () {
         }).to.not.throw();
     });
 
-    it('should error out if v1 collection is passed', function (done) {
```
kunagpal: Skip this test with a todo instead of removing it altogether.

lib/run/index.js
```diff
@@ -54,8 +54,8 @@ var fs = require('fs'),
      * @type {Object}
      */
     knownReporterErrorMessages = {
-        html: '  run `npm install newman-reporter-html`',
-        teamcity: '  run `npm install newman-reporter-teamcity`'
+        html: '  run `npm install newman-reporter-html`\n',
```
shamasis: lol ... i spotted this sneaky out of context change in this PR. don't make a habit of it. :-p

PR #1670, opened 2018-08-08
Throw error for invalid color values

codenirvana:


bin/util.js
```diff
@@ -57,6 +57,14 @@ module.exports = {
          */
         csvParse: (list) => {
             return _.split(list, ',');
+        },
+
+        color: (value) => {
+            if (!(/^(auto|on|off)$/).test(value)) {
+                throw new Error(`invalid value \`${value}\` for option color`);
```
kunagpal: Also include the expected values in the error message.

bin/util.js
```diff
@@ -57,6 +57,14 @@ module.exports = {
          */
         csvParse: (list) => {
             return _.split(list, ',');
+        },
+
+        color: (value) => {
```
shamasis: This is not type `color`. It is `colorOptions` - color would be hex code, colour names, etc

PR #1673, opened 2018-08-09
Fix cli reporter not showing [errored] on request errors

iamwillbar:
The check for err prevented the message being displayed.

### DIFF
--- a/lib/reporters/cli/index.js
+++ b/lib/reporters/cli/index.js
@@ -151,15 +151,18 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
 
     // output the response code, reason and time
     emitter.on('request', function (err, o) {
-        if (err) { return; }
+        if (err) {
+            print.lf(colors.red('[errored]'));
+
+            return;
+        }
 
         var size = o.response && o.response.size();
 
         size = size && (size.header || 0) + (size.body || 0) || 0;
 
-        err ? print.lf(colors.red('[errored]')) :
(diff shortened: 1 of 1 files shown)


lib/reporters/cli/index.js
```diff
@@ -151,15 +151,18 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
 
     // output the response code, reason and time
     emitter.on('request', function (err, o) {
-        if (err) { return; }
+        if (err) {
+            print.lf(colors.red('[errored]'));
```
kunagpal: Could you also add the error message after `[errored]`? That will make debugging a whole lot easier 😄

iamwillbar: I was trying to keep it consistent with what was there before, but that improvement makes sense. I'll send an update shortly.

kunagpal:
@iamwillbar I did a quick visual check with these changes, and the actual error message ends up overflowing. Could you adjust the error printing as follows?
```js
if (err) {
    print.lf(colors.red('[errored]'));
    print.lf(colors.red(err.message));
}
```

shamasis:
@iamwillbar - We really appreciate the PR. ❤️

iamwillbar:
@kunagpal I ended up indenting the error message to make it clearer that it relates to that request:

```
→ Non-Existent Host
  POST https://non-existent.local [errored]
     getaddrinfo ENOTFOUND non-existent.local non-existent.local:443
  2. Status code is 200
  3. No errors returned
```

kunagpal:
@iamwillbar This has been released with Newman v4.0.2. Thanks for contributing! 😄

PR #1717, opened 2018-09-15
Create new `newman_alpine` image based on latest alpine version

### LINKED ISSUE #1715: Security Issue with Alpine-based Docker Images
Hello there!

The current release of Alpine 3.8.1 contains an important security fix for the apk package manager: https://alpinelinux.org/posts/Alpine-3.8.1-released.html

Are there any plans to upgrade the alpine based newman image `newman_alpine33` to the latest version?

Maybe I could provide a pull request updating the image. A first round of testing revealed no compatibility breaking changes introduced by Alpine 3.8. However I would need some information about the best way to provide the image: as a new `newman_alpine38`? IMHO a general image not tied to a specific OS version could be provided as well, i.e. replacing `newman_alpine33` with an image called `newman_alpine`. This allows you to transparantly upgrade the alpine version without requiring actions from users.

namelessvoid:
This PR introduces a new `newman_alpine` image which allows to transparently update to newer alpine versions. Fixes #1715.


docker/images/alpine/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+# Set environment variables
+ENV LC_ALL="en_US.UTF-8" LANG="en_US.UTF-8" LANGUAGE="en_US.UTF-8" ALPINE_NODE_REPO="oznu/alpine-node"
+
+# Bail out early if NODE_VERSION is not provided
+RUN if [ ! $(echo $NEWMAN_VERSION | grep -oE "^[0-9]+\.[0-9]+\.[0-9]+$") ]; then \
+        echo "\033[0;31mA valid semver Newman version is required in the NEWMAN_VERSION build-arg\033[0m"; \
+        exit 1; \
+    fi && \
+    # Add the latest Alpine Linux package to the repositories list
+    echo -e "http://nl.alpinelinux.org/alpine/v3.8/main\nhttp://nl.alpinelinux.org/alpine/v3.8/community" > /etc/apk/repositories && \
```
namelessvoid: Is manually adding these two repositories mandatory? I removed this line and the image seemed to work just fine (using the test requests from https://www.getpostman.com/collections/8a0c9bc08f062d12dcda).

kunagpal: @namelessvoid This is necessary for Alpine 3.3, we should remove them if the image builds and runs fine 😄

kunagpal: @namelessvoid 
**Update:** I tried building this locally, and there were consistent errors installing dependencies. This could be due to our builds picking up from different cache edge locations. Let's stick with this for consistency.

kunagpal:
@namelessvoid Thanks for contributing to Newman! 😄

namelessvoid:
It's been a pleasure! :smile: 

When Do you plan to push the image to Docker Hub?

kunagpal:
@namelessvoid See https://hub.docker.com/r/postman/newman/tags/ 😄

namelessvoid:
Nice job, thank you very much!

Are all `*-alpine` images based on the `alpine:3.8` image?

kunagpal:
@namelessvoid Yes, the Alpine tags are based on the [Node Docker Alpine image](https://github.com/nodejs/docker-node/blob/336fb229392876a5f0d893436aeccf8c80011eeb/10/alpine/Dockerfile), which in turn is based on `alpine:3.8`.

PR #1751, opened 2018-10-15
chore(package): update editorconfig to version 0.15.2

kunagpal:
Closes #1750


shamasis:
For what joy is Travis failing on this?! 🙄

shamasis:
Looks like integration test. The test isn’t flaky. Then why the fuss. 😕

codenirvana:
@shamasis This was flaky earlier, fixed in #1684.
Looks like we need to reduce the limit of parallel executions.

PR #1823, opened 2018-12-15
Add skipped status to beforeDone assertion summary

matt-ball:
Assertion summaries currently cover 2/3 possibilities for an assertion.

- If `error` is `undefined`, then it's assumed it `passed`
- Vice versa for error

If a test is skipped, it cannot be distinguished from a passed test at this level. This will be really useful for reporters where tests are aggregated/tabled, imagine a third column which would allow you to see what tests have skipped:

image

Whilst the `assertion` event contains the `skipped` status, there doesn't seem to be a clear way to join/aggregate assertions from `assertion` with `beforeDone`, so making `skipped` available within `assertions` on `beforeDone` seems to make most sense.

### DIFF
--- a/lib/run/summary.js
+++ b/lib/run/summary.js
@@ -178,6 +178,7 @@ _.assign(RunSummary, {
 
             (execution.assertions || (execution.assertions = [])).push({
                 assertion: o.assertion,
+                skipped: o.skipped,
                 error: err || undefined
             });
         });


DannyDainton:
Is there a reason why this was not added before this PR, the property comes through on the `assertion` event. The only place I can see that currently being used is in the CLI reporter.

It would be great to make `assertions` on the `summary.run.executions` object consistent.

lib/run/summary.js
```diff
@@ -178,6 +178,7 @@ _.assign(RunSummary, {
 
             (execution.assertions || (execution.assertions = [])).push({
                 assertion: o.assertion,
+                skipped: o.skipped,
```
DannyDainton: I think that this will always be set to `true` even for non skipped tests. It's only a visible property if the test is skipped, if it's not skipped, you wont see the property. 

I think it should be `o.skipped || undefined` to make it false for non skipped tests.

DannyDainton: Testing this change out this morning, as Matt originally had it - I had no skipped tests in the collection and this property was still set to true, when it should have been false.

I've seen that the `skipped` property is not always part of the `assertion` event, it's only added when there is a skipped test. It's doesn't flip between `true`/`false` - it's either there and `true` or just not part of the object.

By adding the `|| undefined` it will show all the time and only be `true` when there is an actual skipped test.

matt-ball: Here's an example of that structure in practice:

```js
{
    cursor: {
        position: 0,
        iteration: 0,
        length: 3,
        cycles: 1,
        empty: false,
        eof: false,
        bof: true,
        cr: false,
        ref: '96ec040b-15fc-49ec-9c68-29e703efecdc',
        httpRequestId: '1076041b-64f6-4135-a8fc-722e8ad44fa9',
        scriptId: '62c32b6c-d6ef-4be3-a4f1-33d04ad5da25',
        execution: '941ca74b-2027-471b-82be-266f81e49ba8'
    },
    assertion: 'First Status code is 200',
    skipped: false,
    error: null,
    item: PostmanItem {
        _: {
            postman_id: '41acd1b4-04d2-41a3-bf7a-8580b80c1b46'
        },
        id: '41acd1b4-04d2-41a3-bf7a-8580b80c1b46',
        name: 'First',
        request: PostmanRequest {
            url: [Object],
            headers: [Object],
            method: 'POST',
            body: [Object]
        },
        responses: PostmanPropertyList {
            members: [],
            reference: {},
            Type: [Object]
        },
        events: PostmanEventList {
            members: [Array],
            reference: {},
            Type: [Object]
        }
    }
}
```

DannyDainton: My apologies for confusing things, not really sure how I got to the conclusion I did :/ 

Awesome to see the change has been approved! :+1: 

Any timeline on when the newman version will be out? :)

kunagpal: @DannyDainton Soon after the builds are green: https://travis-ci.org/postmanlabs/newman/builds 😄

DannyDainton: You guys are freaking awesome!! Sorry for confusing things and thank you to Matt for raising this after our chat last night. I'll just stick to hacking about and being in awe of you all :)

PR #1828, opened 2018-12-19
Make JUnit reporter output more friendly for Azure DevOps

sandermvanvliet:
In our set up of running Postman tests using Newman in our Azure DevOps release pipeline we've seen that the xml file containing the test results is missing some information that Azure DevOps uses to populate data in the Tests tab for a release.

Based on [this document](https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/test/publish-test-results?view=vsts&tabs=yaml#result-formats-mapping) I've added the following attributes to the test output file:

* `timestamp` attribute on `testsuite`. Taken from `summary.run.timings.started`
* `classname` attribute on `testsuite`. Taken from `attributes.name.value` which corresponds to the Postman item name
* `message` attribute on `failure` element in a `testcase`.

These attributes are then used by Azure DevOps to populate:

* The `Date started` property
* The `Test file` property
* The `Error message` property, for failed tests. This surfaces the error message quicker than the inner stack trace details.


kunagpal:
Hey @sandermvanvliet, thanks for creating this pull request. We'll start reviewing your code changes shortly 😄

sandermvanvliet:
I just noticed (late I know) that with Newman v4 the HTML reporter is in it's own package now. Is the idea to do that for the JUnit reporter as well? If so, is there already an effort in that direction that I need to push this to instead?

codenirvana:
@sandermvanvliet No, JUnit reporter is still an inbuilt reporter.

lib/reporters/junit/index.js
```diff
@@ -34,6 +36,9 @@ JunitReporter = function (newman, reporterOptions) {
             accumulator[execution.id].push(execution);
         }, {});
 
+        date = new Date(_.get(newman, 'summary.run.timings.started'));
```
kunagpal: `date` is not needed, `timestamp` can be assigned directly.

sandermvanvliet: Done.

kunagpal:
@sandermvanvliet Thanks for your contribution! 😄

PR #1840, opened 2018-12-27
Fix issue causing incomplete junit report to be generated for nested requests 

### LINKED ISSUE #1461: Incomplete junit report when having nested requests/tests
1. Newman Version (can be found via `newman -v`): 
3.9.3
macOS 10.13.3
 3. Are you using Newman as a library, or via the CLI? 
CLI
 3. Did you encounter this recently, or has this bug always been there: -
 4. Expected behaviour: 
The junit reporter should include all assertions in a report (including those of nested requests). The CLI reporter and the postman app are able to show all assertions.
`newman run nestedRequests.postman_collection.json -r cli,junit`
https://www.getpostman.com/collections/71794c3e8d7dcc3c1782
 7. Screenshots (if applicable): -

1. Download &  run the collection (see above 7. & 6.)
2. Check the junit result stored in `./newman`

Details:
For a project I have to execute a POST request to queue a job (where I get a UUID as response). Then I have to send a GET request, using the UUID, to retrieve more details on the queued job.
I don't want to copy/paste the same GET request over and over in my collection and therefore I have implemented a nested request within the test script of a folder (and therefore for multiple requests). This works perfectly but when I try to get junit reports via newman, the xml does not include the assertions for nested requests.

The sample collection has two requests:
- The first is a simple request having a `pm.test()` defined as test-script.
- The second has a nested request `pm.sendRequest()` and a related `pm.test()` as test-script.

ankitsiam:
When given a collection with nested requests i.e requests in test scripts, Junit reporter is unable to generate a complete report. 

Fixes #1461

### DIFF
--- a/lib/reporters/junit/index.js
+++ b/lib/reporters/junit/index.js
@@ -31,8 +31,8 @@ JunitReporter = function (newman, reporterOptions) {
         root.att('tests', _.get(newman, 'summary.run.stats.tests.total', 'unknown'));
 
         cache = _.transform(report, function (accumulator, execution) {
-            accumulator[execution.id] = accumulator[execution.id] || [];
-            accumulator[execution.id].push(execution);
+            accumulator[execution.item.id] = accumulator[execution.id] || [];
+            accumulator[execution.item.id].push(execution);
         }, {});
 
         timestamp = new Date(_.get(newman, 'summary.run.timings.started')).toISOString();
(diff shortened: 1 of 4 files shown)


lib/reporters/junit/index.js
```diff
@@ -31,8 +31,8 @@ JunitReporter = function (newman, reporterOptions) {
         root.att('tests', _.get(newman, 'summary.run.stats.tests.total', 'unknown'));
 
         cache = _.transform(report, function (accumulator, execution) {
-            accumulator[execution.id] = accumulator[execution.id] || [];
-            accumulator[execution.id].push(execution);
+            accumulator[execution.item.id] = accumulator[execution.id] || [];
```
codenirvana: @kunagpal Isn't this assignment wrong now?
`accumulator[execution.id]` will be always null.

codenirvana: `execution.id` is actually `item.id` 🤔
https://github.com/postmanlabs/newman/blob/develop/lib/run/summary.js#L152

PR #1887, opened 2019-02-12
Added --verbose option to show detailed timing split of request

vikiCoder:
### Without --verbose:
without verbose

### With --verbose:
capture


vikiCoder:
Should I align text in timing information to _center_ or to _right_? I think _right_ aligned text in table looks more standard to command-line tables. Below is a semple of both...

### align: Center - 
center

### align: Right - 
right

lib/reporters/cli/cli-utils.js
```diff
@@ -46,6 +46,29 @@ cliUtils = {
         'right-right': ''
     },
 
+    /**
+     * A set of CLI table symbols for compact borderless table.
+     *
+     * @type {Object}
+     */
+    cliTableTemplate_Compact: {
```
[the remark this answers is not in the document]
vikiCoder: To print borderless table. There is cliTableTemplate_Blank charset. But using that set adds a blank line between two rows, which we don't want. So I created a new set.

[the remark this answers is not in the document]
kunagpal: Okay, change the variable name to be camelCased.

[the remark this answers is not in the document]
kunagpal: Also, derive this from the `cliTableTemplate_Blank` object to avoid unwarranted duplication.

lib/util.js
```diff
@@ -64,9 +64,47 @@ util = {
...
 
+    /**
+     * Returns the time  object with all values in largest time unit possible as strings.
+     *
+     * @param {Object} obj - {event1: time1, event2: time2, ...} (time in milliseconds)
+     * @returns {Object} - {event1: time1, event2: time2, ...} (time in string with appropriate unit)
+     */
+    beautifyTime: function (obj) {
+        const intervals = {
+            m: 60000000,
```
kunagpal: Use the scientific notation for better readability.

lib/reporters/cli/index.js
```diff
@@ -315,6 +358,42 @@ _.assignIn(PostmanCLIReporter, {
             hAlign: 'left'
         }]);
 
+        timings.waitAverage && summaryTable.push([{
```
kunagpal: Most of the code in this section is duplicated, can be compacted with a list and a loop 😄

lib/run/summary.js
```diff
@@ -227,7 +270,7 @@ _.assign(RunSummary, {
 
             var size = _.isFunction(o.response.size) && o.response.size(),
                 time = o.response.responseTime,
-
+                timingPhases = sdk.Response.timingPhases(o.response.timings),
```
kunagpal: @codenirvana This function should be on the response instance as well.

lib/reporters/cli/index.js
```diff
@@ -180,7 +180,7 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
         // print timing info of the request
         if (options.verbose && timingPhases) {
             timingTable = new Table({
-                chars: cliUtils.cliTableTemplate_Compact,
+                chars: _.assign({}, cliUtils.cliTableTemplate_Blank, { mid: '', middle: ' ' }),
```
kunagpal: `_.defaults({ mid: '', middle: '' }, cliUtils.cliTableTemplate_Blank)`

shamasis:
@DannyDainton - I think this might interest you. 😛

DannyDainton:
That's awesome!! Obviously, going to be displaying this on the reports now!! Great work!! 🏆

shamasis:
Hopefully next release. But wait for a while - we might change the data structure slightly to correct for inclusion of processing overhead in the stats too. 

We have seen so many tools exclude the processing timings. Wanted to ensure we expose that bit too.

DannyDainton:
Sure. I will hold off until its stable. 😁

I want to try and allow people to use my reporter (sorry it's not Postman's version 😁) to expose all that they can - Or at least have the option too if they make their own templates.

codenirvana:
@DannyDainton Added in Newman v4.4.0. 🎉

DannyDainton:
I've been having a look already, awesome work! Just trying to think how best to display that data rather than just dump in on the report. 

Appreciate being kept in the loop. :) I'm very nearly part of the Postman team....Days to go :)

Issue #1931, opened 2019-03-09
Junit report doesn't pass validation against schema in xunit Jenkins plugin


donotello:
1. Newman Version (can be found via `newman -v`): `4.4.0`
 2. OS details (type, version, and architecture): `Ubuntu 18.04.2 LTS`
 3. Are you using Newman as a library, or via the CLI? `CLI`
 4. Did you encounter this recently, or has this bug always been there: introduced in 4.3.1 (#1828)
 5. Expected behaviour: junit report generated by newman is successfully validated against [Ant junit](https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd) and [Maven surefire](http://maven.apache.org/surefire/maven-surefire-plugin/xsd/surefire-test-report.xsd) schemas
 6. Command / script used to run Newman: `newman run single-request-failing.json -r junit --reporter-junit-export test.xml`
 7. Sample collection, and auxiliary files (minus the sensitive details): you can use any collection e.g. [single-request-failing.json](https://github.com/postmanlabs/newman/blob/develop/test/fixtures/run/single-request-failing.json)

This issue breaks our integration of newman with [Jenkins xUnit plugin](https://wiki.jenkins.io/display/JENKINS/xUnit+Plugin):

```
INFO: [JUnit] - 1 test report file(s) were found with the pattern 'newman-output/xml/Newman-Output.xml' relative to '/scm/Checkout/develop' for the testing framework 'JUnit'.
WARNING: The file '/scm/Checkout/develop/newman-output/xml/Newman-Output.xml' is an invalid file.
WARNING: At line 3 of file:/scm/Checkout/develop/newman-output/xml/Newman-Output.xml:cvc-complex-type.3.2.2: Attribute 'classname' is not allowed to appear in element 'testsuite'.
...
```

codenirvana:
@donotello This is fixed in Newman v4.4.1. Thanks for contributing!

Issue #1678, opened 2018-08-14
JUnit reporter: Invalid content was found starting with element 'error'


nblock:
1. Newman Version (can be found via `newman -v`): 4.0.2
 2. OS details (type, version, and architecture): Jenkins 2.212.2, xUnit 2.1.0
 3. Are you using Newman as a library, or via the CLI: CLI
 3. Did you encounter this recently, or has this bug always been there: Just started using it
 4. Expected behaviour: xUnit should be able to parse the XML file.
 5. Command / script used to run Newman: `newman run ... --reporters cli,junit --reporter-junit-export $WORKSPACE/postman.xml`

Error message from xUnit during parsing:
```
WARNING: The file '/home/jenkins/workspace/jobname/postman.xml' is an invalid file.
WARNING: At line 4 of file:/home/jenkins/workspace/jobname/postman.xml:cvc-complex-type.2.4.a: Invalid content was found starting with element 'error'. One of '{testsuite, properties, testcase, system-out, system-err}' is expected.
```

The XML file generated by running newman:
```
<?xml version="1.0" encoding="UTF-8"?>
<testsuites name="Call all login pages" tests="109" time="38.696">
  <testsuite name="open login page" id="7f72e270-d492-452e-a41e-d1f27a2d5de9" tests="4" failures="5" errors="2" time="0.355">
    <error>
      <![CDATA[Iteration: 9
RequestError: Error: write EPROTO 140111312403328:error:140770FC:SSL routines:SSL23_GET_SERVER_HELLO:unknown protocol:../deps/openssl/openssl/ssl/s23_clnt.c:827:

    at WriteWrap.afterWrite [as oncomplete] (net.js:844:14)

---
testScriptError: Cannot read property 'has' of undefined
---
]]>
    </error>
    <testcase name="Body does NOT contain string maintenance mode" time="0.337"/>
    <testcase name="Status code is 403" time="0.337"/>
    <testcase name="Response time is less than 2500ms" time="0.337"/>
    <testcase name="Body contains string user login" time="0.337"/>
  </testsuite>
</testsuites>
```

I guess the text inside the `error` node should be wrapped inside `system-err`

carlowahlstedt:
Looks like someone found [this issue](https://github.com/carlowahlstedt/NewmanPostman_VSTS_Task/issues/42#event-1992211269) using Newman via AzureDevOps.

donotello:
[...]

PR #1932, opened 2019-03-10
Fix for #1931: remove classname from suite and it to both passed and failed tests.

donotello:
This is a fix for #1931 and #1678. I didn't separated those two since they are both about passing the schema validation in Jenkins xUnit plugin. 

### Remove classname from suite and add it to both passed and failed tests. (#1931)

As part of #1828 classname attribute was added to testsuite. This causes [Jenkins xunit plugin](https://wiki.jenkins.io/display/JENKINS/xUnit+Plugin) to fail to process reports generated by newman.

The following links proves that classname should be added only for testcase:

* https://docs.microsoft.com/en-us/azure/devops/pipelines/tasks/test/publish-test-results?view=azure-devops&tabs=yaml&viewFallbackFrom=vsts#result-formats-mapping
* https://github.com/windyroad/JUnit-Schema/blob/master/JUnit.xsd
* http://maven.apache.org/surefire/maven-surefire-plugin/xsd/surefire-test-report.xsd

At the same time classname was added only to failing tests which causes incorrect calculation of age for the case by the same xunit plugin. In order to fix this the classname attribute was added to both failed and passed tests.

The classname now also generated from collection name based on the [Java identifiers rules](https://docs.oracle.com/javase/specs/jls/se8/html/jls-3.html#jls-3.8). This is better than constant name cause allows to visually separate tests from different collections if we run several of them as part of a single Job in Jenkins so it will looks like this:

### Rename `error` element to `system-err` (#1678)

In case of runtime javascript errors printed to stderr we had `error` tag in xml report that also called issues with Jenkins xunit plugin. This tag is now replaced with `system-err` and it passes validation.

### Validation

I've validated those changes with local Jenkins + xUnit plugin 2.3.3 with the reports from the following collections:

* [failed-request.json](https://github.com/postmanlabs/newman/blob/develop/test/fixtures/run/failed-request.json)
* [nested-requests-fail-report-test.json](https://github.com/postmanlabs/newman/blob/develop/test/fixtures/run/nested-requests-fail-report-test.json)
* [nested-requests-report-test.json](https://github.com/postmanlabs/newman/blob/develop/test/fixtures/run/nested-requests-report-test.json)
* [newman-report-test.json](https://github.com/postmanlabs/newman/blob/develop/test/fixtures/run/newman-report-test.json)
* [response-bodies.json](https://github.com/postmanlabs/newman/blob/develop/test/fixtures/run/response-bodies.json)
* [single-get-request.json](https://github.com/postmanlabs/newman/blob/develop/test/fixtures/run/single-get-request.json)
* [singe-request-failing.json](https://github.com/postmanlabs/newman/blob/develop/test/fixtures/run/singe-request-failing.json)
* [undefined-test-checks.json](https://github.com/postmanlabs/newman/blob/develop/test/fixtures/run/undefined-test-checks.json)

```
Started by user donotello
Building in workspace /var/jenkins_home/workspace/xunit
INFO: Starting to record.
INFO: Processing JUnit
INFO: [JUnit] - 8 test report file(s) were found with the pattern '*.xml' relative to '/var/jenkins_home/workspace/xunit' for the testing framework 'JUnit'.
INFO: Setting the build status to SUCCESS
INFO: Stopping recording.
Build step 'Process xUnit test result report' changed build result to SUCCESS
Finished: SUCCESS
```

image

image

P.S. I was thinking about adding ant and maven junit schemas to repository and add validation of reports against those schemas but existing modules for xsd schema validation will add some dependency (e.g. Java). Also looks like validation in xunit plugin is not really strict and strict validation will require much more changes.


donotello:
@sandermvanvliet, could you please also review this pull request since it partially removes part of the code you've added in #1828.

lib/reporters/junit/index.js
```diff
@@ -20,12 +20,15 @@ JunitReporter = function (newman, reporterOptions) {
             root,
             testSuitesExecutionTime = 0,
             executionTime = 0,
-            timestamp;
+            timestamp,
+            classname = _.camelCase(collection.name).replace(/\W/g, '');
```
kunagpal: @donotello This statement is redundant, see the comment below.

lib/reporters/junit/index.js
```diff
@@ -20,12 +20,15 @@ JunitReporter = function (newman, reporterOptions) {
...
 
         if (!report) {
             return;
         }
 
+        classname = classname[0].toUpperCase() + classname.slice(1);
```
kunagpal: Since we're already bailing out in the `if` block above, this statement can be re-written as:

```suggestion
        classname = _.upperFirst(_.camelCase(classname));
```

lib/reporters/junit/index.js
donotello: Rewritten, thank you.

lib/reporters/junit/index.js
donotello: Thank you. Rewritten. I've left replacement of non-word characters so because implementation of `camelCase` in lodash won't filter some characters that can't be part of java class name.

kunagpal:
@donotello L.G.T.M, thanks for contributing to Newman! 😄

PR #2003, opened 2019-05-13
develop < fix-json-silent

mkhatuntcov-korewireless:
I would like to have an option silent for JSON reporter.

basically replicate this logic:
https://github.com/postmanlabs/newman/blob/develop/lib/reporters/cli/index.js#L70


codenirvana:
I don't think `silent` flag is required for the JSON reporter.
CLI being the default reporter needed some way to completely turn all reporters off.

PR #2043, opened 2019-06-24
Fixed CSV parsing for fields containing quotes and files with BOM characters

vikiCoder:
**Issues:**
- User can't use double quotes (`"`) inside an unquoted field in CSV file without escaping them
https://github.com/postmanlabs/postman-app-support/issues/6764

**Resolution:**
- Use `relax` option of `csv-parse` library to allow using double quotes `"` without escaping inside an unquoted field in CSV file. However, the quoted field still need to escape `"` using `\` inside the field value.


package.json
```diff
@@ -45,7 +45,7 @@
     "cli-table3": "0.5.1",
     "colors": "1.3.3",
     "commander": "2.20.0",
-    "csv-parse": "4.0.1",
+    "csv-parse": "4.4.3",
```
codenirvana: Drops support for Node v6. Can't do it now.

lib/run/options.js
```diff
@@ -285,7 +285,9 @@ var _ = require('lodash'),
                             columns: true,
                             escape: '\\',
                             cast: csvAutoParse,
-                            trim: true
+                            trim: true,
+                            relax: true, // allow using quotes without escaping inside unquoted string
+                            bom: true // exclude the byte order mark (BOM) from the CSV input if present
```
codenirvana: Remove this option, as this requires a dependency update.

Issue #2069, opened 2019-07-24
Document all events (and their args) emitted by the `newman` lib


kaustavdm:
Currently, there is only a list of events documented for newman in the [README.md](https://github.com/postmanlabs/newman#newmanrunevents). These events need to be documented as:

## event name

Event Description

args           |   description
-----------|--------------------
`opts`       | This is an option
`opts.foo` | A property of the `opts` object

malvikach:
Documentation can be found [here](https://github.com/postmanlabs/newman/wiki/Newman-Run-Events)

PR #2074, opened 2019-07-29
GH-2069-added description to events table for events raised by newman

arnitkun:
Updated the events table that contains description of the arguments for the events raised by Newman.


codenirvana:
Need a better way to represent both event and arguments descriptions.

PR #2147, opened 2019-10-08
Adds example for `options.globals` parameters

filet0fish:
Adds documentation for the object structure expected by `options.globals`.  This was requested in [this post](https://community.getpostman.com/t/how-to-specify-globals-directly-using-newman-run-in-nodejs/490).


codenirvana:
Thanks for your contribution!

Maybe we should consider adding examples for every option in a collapsable section.
But, at this point, I don't feel an example of a `globals` options is needed as in most cases it points to a file exported from Postman.

Another reason is, the existing table representation for documenting options is not suitable for including long-text examples like this.

We will try to fix these documentation issues as soon as possible!

Issue #1839, opened 2018-12-27
Multiple ssl key and cert in the same test


dtalex:
Hi,
I use newman latest verssion of newman via cli
I have the need to send https requests to different servers supporting mutual authentication in the same test simulating different users.
As of now in newman I can only set one couple key/cert.

I implemented it Q&D for my use here [https://github.com/dtalex/newman/tree/myMod](url)

Basically here 
```javascript
 certificates: options.sslClientCert && new sdk.CertificateList({}, [{
                name: 'client-cert',
                matches: [sdk.UrlMatchPattern.MATCH_ALL_URLS],
                key: { src: options.sslClientKey },
                cert: { src: options.sslClientCert },
                passphrase: options.sslClientPassphrase
            }])
```

I feed a configuration file done as follows
```json
[{
	"name": "identity1-cert",
	"matches": "https://server1.com/*",
	"key": {
		"src": "C:/a/b/identity1.key"
	},
	"cert": {
		"src": "C:/a/b/identity1.cer"
	}
},
{
	"name": "identity2-cert",
	"matches": "https://server2.com/*",
	"key": {
		"src": "C:/a/b/identity2.key"
	},
	"cert": {
		"src": "C:/a/b/identity2.cer"
	},
	"passphrase": "qwer1234"
}]
```

Is it possible to see it implemented in future releases?
(Probably I had to post it in Pull Requests but it is too "ugly" to be accepted)
Regards,
Ale

sandeepkumar-aitha:
I too need this feature.

sandeepkumar-aitha:
I too need this feature

codenirvana:
`--ssl-client-cert-list` option is added in Newman v4.6.0 🎉

s4ik4t:
Do not see any mention for pfx files, is that supported too?

PR #2178, opened 2019-11-21
add multiple ssl client certificates support

chengchengmu:
Hello @shamasis and @codenirvana ,

 # Description
Postman already supports a client certificate per domain.
This PR implements a similar feature:
**one can provide multiple ssl client certificate for the same collection run: a different ssl client certificate per url.**

The original way to pass ssl client certificate is still supported for backward compatibility and can be handy for single certificate use cases.

Related issue: https://github.com/postmanlabs/newman/issues/1839

# Example
Example of ssl client certificate configuration file (can be found in  examples/ssl-client-cert-list.json)
```
[
    {
        "name": "domain1",
        "matches": ["https://test.domain1.com/*", "https://www.domain1/*"],
        "key": {"src": "./client.domain1.key"},
        "cert": {"src": "./client.domain1.crt"},
        "passphrase": "changeme"
    },
    {
        "name": "domain2",
        "matches": ["https://domain2.com/*"],
        "key": {"src": "./client.domain2.key"},
        "cert": {"src": "./client.domain2.crt"},
        "passphrase": "changeme"
    }
]
```

Example of command:
```
newman.js run  --ssl-client-certs sslClientCertConfig.json  collection.json
```

# Changes
[x] Documentation update
[x] Tests
[x] Feature


modusemerson:
Hi when is this going to be publish? this is a really helpful tool pls approve the merge

chengchengmu:
hello @modusemerson , hopefully @shamasis or @codenirvana can have a look when they get the chance.

For now we are using a forked version, which is not great as we need to actively keep it up to date with upstream

README.md
```diff
@@ -217,6 +217,14 @@ Client certificates are an alternative to traditional authentication mechanisms.
 
 Newman supports SSL client certificates, via the following CLI options (available with Newman `v3` style `run` only):
 
+#### Use a ssl client certificates configuration file (supports multiple certificates per run)
+
+- --ssl-client-certs<br/>
```
codenirvana: Let's call this `--ssl-client-cert-list`

README.md
```diff
@@ -217,6 +217,14 @@ Client certificates are an alternative to traditional authentication mechanisms.
...
+The path to the ssl client certs configuration file (json format). See [examples/multiple-ssl-client-certs.json](https://github.com/postmanlabs/newman/blob/develop/examples/multiple-ssl-client-certs.json)
```
codenirvana: We should add an example (examples/ssl-client-cert-list.json) as part of this PR.

README.md
```diff
@@ -217,6 +217,14 @@ Client certificates are an alternative to traditional authentication mechanisms.
...
+This option allows multiple ssl client certificiates and different ssl client certificate per url or hostname.
+This option and --ssl-client-cert, --ssl-client-key and --ssl-client-passphrase are exclusives.
```
codenirvana: How about merging individual cert config into the cert list?
And the individual certs take priority over the same config present in the list (works as overrides).

bin/newman.js
```diff
@@ -52,11 +52,15 @@ program
...
+    .option('--ssl-client-certs <path>',
+        'Specify the path to the Client SSL certificates configuration file (json format).' +
+        '\nThis option different ssl client certificate per url/hostname.' +
+        '\nThis option and --ssl-client-cert, --ssl-client-key and --ssl-client-passphrase are exclusives.')
     .option('--ssl-client-cert <path>',
         'Specify the path to the Client SSL certificate. Supports .cert and .pfx files.')
     .option('--ssl-client-key <path>',
         'Specify the path to the Client SSL key (not needed for .pfx files)')
-    .option('--ssl-client-passphrase <path>',
+    .option('--ssl-client-passphrase <passphrase>',
```
codenirvana: 👍

README.md
chengchengmu: @codenirvana I had the same idea while implementing and testing this feature.
However the individual certificate would work differently than the list:
the individual cert does not have url prefix parameter, how to determine the priority ?

We can decide to use the individual certificate as:
1. the default certificate, i.e. for any url not matched by the url prefixes defined in the list. 
2. Or to use it as the most priority certificate, however there is no url prefix associated to it, so it would be applied all the time: it's still possible by not passing the certificate list.

I actually tried to implement solution 1. and it's not trivial:
Currently if we set url prefix `*` to the individual cert to make it the default, it actually gets matched first even if there is a more specific url prefix.

So I think if we want to implement 1. we will need to change the priority order of url prefixes matching.
For 2, we can just omit the certificate list.

To make it simpler for now, the 2 options are exclusives: only one can be used

README.md
chengchengmu: Good catch. Will add this file.

chengchengmu:
Thank you @codenirvana for your review and all good for the delay. Really appreciate your suggestions and I share your opinion. As explained earlier the merge of the two options might not be trivial, would you be ok to look at it in a separate time, and leave it to a more qualified engineer ?

I have just made changes to add an example of ssl client cert list file and also rename the option (great suggestion).
Please let me know if you prefer squashing the 2 commits into 1

codenirvana:
@chengchengmu You are right. I think the individual cert config can be used as a fallback if nothing is matched from the certificate list config.

@coditva will take this forward.

chengchengmu:
@coditva thanks for taking this forward

@codenirvana the CI test failed the "should be handled correctly when breached" in nodejs, could it be a flaky one ?
Would love your review again as the PR has been updated

jeffplesko:
Anymore updates on this change request?  We are at a point where we need this immediately for our automation.  Please provide any updates as it appears to be 1 month since the last integration test/merge was attempted.

chengchengmu:
Hello @codenirvana  @coditva ,

FYI just rebased this branch on `develop`.
The code coverage report shows a file that has been not touched in this PR: lib/reporters/cli/cli-utils.js 

To prevent people using forks of newman with that feature, would you mind considering merging this PR before the improvement (fallback if nothing is matched from the certificate list config.) is implemented ?

coditva:
@chengchengmu Thanks for your changes! I have added the fallback to the individual cert option. I could not push to your branch, so I created a new Pull Request here: #2235. Please let me know if you have any comments! :)

PR #2218, opened 2020-02-01
Update README.md

jcald1:


README.md
```diff
@@ -94,9 +94,9 @@ const newman = require('newman'); // require newman in your project
 newman.run({
     collection: require('./sample-collection.json'),
     reporters: 'cli'
-}, function (err) {
+}, function (err, data) {
 	if (err) { throw err; }
-    console.log('collection run complete!');
+    console.log(`collection run complete! Summary Run Data: ${data}`);
```
codenirvana: the second argument `data` is an object and using it with string template literal will convert it to `[object Object]`.
So, here it will log: `collection run complete! Summary Run Data: [object Object]` 😕

jcald1: @codenirvana, resolved.  I also updated the documentation for running newman as a library.

jcald1:
@codenirvana, updated per your review. I also updated the "as a library" section.

README.md
```diff
@@ -94,12 +94,14 @@ const newman = require('newman'); // require newman in your project
 newman.run({
     collection: require('./sample-collection.json'),
     reporters: 'cli'
-}, function (err) {
+}, function (err, data) {
 	if (err) { throw err; }
-    console.log('collection run complete!');
+    console.log('collection run complete! Summary Run Data', data);
```
codenirvana:
```suggestion
    console.log('collection run complete!');
```

I suggest not including this in the getting started snippet since the `data` object includes all the execution details and in-general very long JSON text. Which I feel will be confusing for the new users.

README.md
```diff
@@ -94,12 +94,14 @@ const newman = require('newman'); // require newman in your project
...
 });
 ```
 
+The collection may either be a URL or a path to a local JSON file.  If it is a URL, you must provide the `postmanApiKey` property.
```
codenirvana: Well, you can give a path to the collection file or any URL where the collection JSON is hosted.
`postmanApiKey` is only needed if fetched via Postman API.

codenirvana (commented):
We've decided to not document the `postmanApiKey` option because it will be deprecated and replaced with the new `newman login` command.

PR #2231, opened 2020-02-22
Adding files to package.json

### LINKED ISSUE #2026: Add "files" to package.json
Using Newman 4.5.0, the eslint-plugin-node recommended ruleset flags `require('newman')` as an unpublished module because the package.json file lacks a [`files` attribute](https://docs.npmjs.com/files/package.json#files). This requires specific eslint configuration to allow newman to pass linting. This seems like unnecessary friction. Refer to the [node/no-unpublished-require rule](https://github.com/mysticatea/eslint-plugin-node/blob/master/docs/rules/no-unpublished-require.md) for additional information. This will also prevent future instances of #245.

Shivansh2407:
Fixes #2026
I have included all the files that need to be included while publishing the package .


Shivansh2407:
Please let me know why are these synk and CI tests failing.

package.json
```diff
@@ -22,6 +22,9 @@
     "newman": "./bin/newman.js"
   },
   "preferGlobal": true,
+  "files": [
+		"*",
```
ZebraFlesh: I think strict JSON parsing doesn't like this trailing comma and that's why the checks are failing.

Shivansh2407: Thanks . Done .

Shivansh2407:
@ZebraFlesh  Please review and merge this 👍

Shivansh2407:
@ZebraFlesh do you know all can merge this PR?

Shivansh2407:
@codenirvana Please review and merge this 👍

ZebraFlesh:
@codenirvana `files` is a whitelist; `.npmignore` is a blacklist. There's quite a bit of extra junk in the published package that is irrelevant to consumers, so the existing blacklist approach has proven ineffective. I think it might make sense in the context of this PR to also delete the `.npmignore` file in order to create a consistent approach.

codenirvana:
@ZebraFlesh Since the two approaches (`files` and `. npmignore`)  are opposite to each other, there are different reasons to pick one over the other.

In my opinion, it makes more sense to choose the `files`  approach for the packages using some kind of transpilation or have distribution (`dist`) directory.

However, for packages like Newman, which can be used as both CLI and library, using the `files` approach will be very dangerous as with exclusion, new files are still published automatically, but with inclusion, you may end up publishing a broken package because you forgot to include the new files.

I know various workaround/pre-publish tests can be used to verify the contents of the published package but let us not get into the debate of choosing one approach over the other.

Finally, I agree with you that some files (primarily `docker/` and `test/`) are unnecessarily getting published so, to fix that let's just add those to the existing `.npmignore` list.

PR #2235, opened 2020-02-24
Add option `--ssl-client-cert-list` to allow setting different certificates for different URLs

coditva:
This PR builds on top of #2178 by @chengchengmu. Thanks for the awesome work! :)
Closes: #1839

### Description
This change adds an option to specify client certificates for specific URLs in the collection. When no certificate is specified in the list which matches a URL, the certificate specified by `--ssl-client-cert` is used (if specified).


chengchengmu:
hello @coditva,

Thanks for the ingenius improvement and thanks @codenirvana  for the suggestion.
It looks very good to me

Issue #1902, opened 2019-02-19
Newman - command-line option to suppress test success details from output


jhardin-aptos:
We're developing data-driven test collections that perform a lot of validations using requests generated from sets of data retrieved from the database. The Newman output from a test run is becoming ... rather unwieldy.

It would be really nice if there was a way to tell newman to not report the details of successful tests, just failed ones.

That would collapse this verbose clutter down to one row in the output:

![image](https://user-images.githubusercontent.com/47017923/52982546-82dd4f00-339b-11e9-8624-a834fe10db56.png)

Note: the _very specific_ test names are due to the Newman HTML output collapsing test results for multiple executions of the same request with different parameters. See issue #1688.

Newman 4.3.1, HTML output
WIN10 CLI

DannyDainton:
@jhardin-aptos Would using a different reporter, that doesn't aggregate the executions help in the short term?

https://www.npmjs.com/package/newman-reporter-htmlextra

jhardin-aptos:
Hm. Potentially.

I also didn't think to take a look at the HTML template options to see if there's a way to create a template that doesn't include the test successes.

Thanks for the suggestion, I will take a look at both possibilities.

DannyDainton:
@jhardin-aptos Just to let you know I've included options in the `htmlextra` reporter that will show you only the `failed` requests and also filtering around the test section. Not sure if this will help with your problem.

jimc404:
@codenirvana I'm working on this

PR #2242, opened 2020-02-26
Feature/no success assertions

jimc404:
This is a fix for #1902. 

When I looked into the codebase, the flag  ```--reporter-cli-no-success-assertions```  was present in passing tests but not in ```bin/newman.js``` (and consequently not in ```newman run -h``` command)

Adding it to ```bin/newman.js``` caused test to fail due to an assertion error. It appears to be a problem with commander.js where they assign default values to flags containing ```no```.  I have added the extra assertion in the test which passes the tests.

This could have been fixed better with some rework but I have left it as is to not potentially break the work-flow of anyone who may be currently using it.

### DIFF
--- a/bin/newman.js
+++ b/bin/newman.js
@@ -29,6 +29,7 @@ program
     .option('-d, --iteration-data <path>', 'Specify a data file to use for iterations (either json or csv).')
     .option('--export-environment <path>', 'Exports the environment to a file after completing the run.')
     .option('--export-globals <path>', 'Specify an output file to dump Globals before exiting.')
+    .option('--reporter-cli-no-success-assertions', 'Show no output on successful assertions')
     .option('--export-collection <path>', 'Specify an output file to save the executed collection')
     .option('--postman-api-key <apiKey>', 'API Key used to load the resources from the Postman API.')
     .option('--delay-request [n]', 'Specify the extent of delay between requests (milliseconds)', util.cast.integer, 0)
--- a/test/unit/cli.test.js
+++ b/test/unit/cli.test.js
@@ -54,8 +54,8 @@ describe('cli parser', function () {
                         envVar: [],
                         folder: [],
                         insecureFileRead: true,
-                        // Removing the following non-comment line causes the test to fail.
-                        // Although the application functioning remains the same in either case.
+                        // Removing the following non-comment line causes the test to fail as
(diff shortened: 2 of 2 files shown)


jimc404:
Can you review this? @codenirvana

jimc404:
Instead of using the reporter-cli options, would it be okay to add a new flag in the app that has the same function? That could make a cleaner fix for the issue.

test/unit/cli.test.js
```diff
@@ -54,6 +54,9 @@ describe('cli parser', function () {
                         envVar: [],
                         folder: [],
                         insecureFileRead: true,
+                        // Removing the following non-comment line causes the test to fail as
```
ankit-m: No need for this comment.

ankit-m:
@jimc404 can you squash the commits and detailed description as to what the commit does?

bin/newman.js
```diff
@@ -29,6 +29,7 @@ program
     .option('-d, --iteration-data <path>', 'Specify a data file to use for iterations (either json or csv).')
     .option('--export-environment <path>', 'Exports the environment to a file after completing the run.')
     .option('--export-globals <path>', 'Specify an output file to dump Globals before exiting.')
+    .option('--reporter-cli-no-success-assertions', 'Show no output on successful assertions')
```
codenirvana: No need to add a reporter specific option.
Newman automatically accumulates all the configurations passed to the `--reporter-` option and pass on to the relevant reporter. More info: https://github.com/postmanlabs/newman#configuring-reporters

codenirvana:
@jimc404 The required change needs to be done in [newman-reporter-html
](https://github.com/postmanlabs/newman-reporter-html)

Also, Newman automatically accumulates all the configurations passed to the --reporter- option and pass on to the relevant reporter. More info: https://github.com/postmanlabs/newman#configuring-reporters

jimc404:
@codenirvana Thanks for pointing out. After checking out the links and taking a good look at the codebase, I realised how embarrassingly bad of a PR this was.

Anyway, I have opened a PR in [newman-reporter-html](https://github.com/postmanlabs/newman-reporter-html) regarding the same issue. Can you help me out in the testing part, How exactly should I place checks to test the HTML output?

PR #2245, opened 2020-03-01
Feature/cli timestamp: Allows the user to have timestamps in the CLI report

### LINKED ISSUE #1998: CLI reporter logging the request start/end times would be very helpful
This is a feature/enhancement request rather than a bug.

The CLI reporter outputs information about each request as it runs, including request duration.
It would also be very helpful to include request start time, and possibly end time too.

For context, when a request in a collection behaves differently to expected, at the moment it's hard to pair up that request with our corresponding service logs, because we have to look in the full time period that the collection was running. Having timestamps on each request would make this troubleshooting process much simpler, because it would be easier to identify the relevant time window.

My current workaround is to also use the JUnit reporter and look through the XML to find the timestamp, but that's not ideal.

Newman Version: 4.4.1
OS: Ubuntu 16.04

jimc404:
Fixes #1998

### What does this PR do?
I have implemented two flags in this PR:

1. ```--reporter-cli-show-timestamp```: This appends a timestamp to each request displaying their start time. It causes the following output:
```
→ A simple GET request
  GET https://postman-echo.com/get?source=newman-sample-github-collection [200 OK, 637B, 1356ms]
```
turn to
```
→ A simple GET request
  Mar 02 2020 03:37:39 GMT+0530  GET https://postman-echo.com/get?source=newman-sample-github-collection [200 OK, 637B, 1520ms]
```
This timestamp also displays the UTC offset of the timezone configured on the system.

The timezone info may not be desired and thus to allow for logging of simpler timestamps there's this flag:
2. ```--reporter-cli-show-timestamp-no-tz```:  This only prints the date and time of the request. The above example with this flag will be:
```
→ A simple GET request
  Mar 02 2020 03:45:56  GET https://postman-echo.com/get?source=newman-sample-github-collection [200 OK, 639B, 1508ms]
```

### Implementation:

- In ```PostmanCLIReporter``` on the event ``` beforeRequest``` check if timestamp options are present.
- If they are present then a string with current system time is fetched. Extract the substring according to the option passed and print it to CLI along with the other data.  

### Testing: 
- For testing of these proposed changes there are 2 new tests added under the file ```test/cli/cli-reporter-show-timestamp.test.js```. 
- They read the CLI output of these features and validate them with regex matches.


jimc404:
@codenirvana Can you provide feedback for this.
TIA

lib/reporters/cli/cli-utils.js
```diff
@@ -75,6 +75,17 @@ cliUtils = {
...
     },
 
+    /**
+     * A helper function which returns a string with the timestamp of current time
+     * (i.e The time when this function is called)
+     *
+     * @returns {String} - A string carrying timestamp
+     */
+    currentTime: function () {
+        return new Date().toString();
```
codenirvana: I feel [toLocaleString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString) works best here.

Check if there's any other helpful instance method available in the Date's [prototype](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date).

codenirvana: We can move this method to the CLI reporter itself if it's not used in multiple places.
Better have fewer utils. 😅

lib/reporters/cli/index.js
```diff
@@ -57,6 +58,9 @@ extractSNR = function (executions) {
  * @param {Boolean=} reporterOptions.noFailures - Boolean flag to turn off failure reporting altogether, if set to true.
  * @param {Boolean=} reporterOptions.noConsole - Boolean flag to turn off console logging, if set to true.
  * @param {Boolean=} reporterOptions.noBanner - Boolean flag to turn off newman banner, if set to true.
+ * @param {Boolean=} reporterOptions.showTimestamp - Boolean flag to log timestamp of each request, if set to true.
+ * @param {Boolean=} reporterOptions.showTimestampNoTz - Boolean flag to log timestamp of each request, if set to true.
```
codenirvana: no sure if this is still needed.

lib/reporters/cli/cli-utils.js
jimc404: > I feel [toLocaleString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString) works best here.
> 
> Check if there's any other helpful instance method available in the Date's [prototype](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date).

So, can we use `toUTCString` and `toLocaleString`. I am of the opinion to allow an option to show UTC time, so that users with newman instances in different timezones can use it out of the box. The flags can be `--reporter-cli-show-timestamp` and `--reporter-cli-show-timestamp-utc` or something similar but maybe shorter. 

What do you think?

lib/reporters/cli/index.js
jimc404: I added this to maintain consistency with the rest of the JSDoc comment.

jimc404: Although being the CLI reporter maybe it's okay to not have it. I'm open to both options as it does not affect the usage either way.

jimc404:
> I feel [toLocaleString](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString) works best here.
> 
> Check if there's any other helpful instance method available in the Date's [prototype](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date).

So, can we use `toUTCString` and `toLocaleString`. I am of the opinion to allow an option to show UTC time, so that users with newman instances in different timezones can use it out of the box. The flags can be `--reporter-cli-show-timestamp` and `--reporter-cli-show-timestamp-utc` or something similar but maybe shorter. 

What do you think? @codenirvana

jimc404:
@codenirvana I made the changes you suggested
- `toLocaleString()` is used to show local time.
- Move the one liner function directly into the timestamp code block
- Removed the 2 JsDoc comments
- Updated README

Please review this whenever you get a chance

README.md
```diff
@@ -425,6 +425,9 @@ such a scenario.
 | CLI Option  | Description       |
 |-------------|-------------------|
 | `--reporter-cli-silent`         | The CLI reporter is internally disabled and you see no output to terminal. |
+
+| `--reporter-cli-show-timestamp` | This prints the local time for each request made. | 
+| `--reporter-cli-show-timestamp-utc` | This prints the UTC time of each request made. |
```
codenirvana: I don't feel any need for this option right now. We can just have a single flag to show the timestamps.

README.md
```diff
@@ -425,6 +425,9 @@ such a scenario.
 | CLI Option  | Description       |
 |-------------|-------------------|
 | `--reporter-cli-silent`         | The CLI reporter is internally disabled and you see no output to terminal. |
+
+| `--reporter-cli-show-timestamp` | This prints the local time for each request made. |
```
codenirvana: shouldn't this be called `--reporter-cli-show-timestamps` 🤔 , thoughts?
Or, can this change be just part of existing `--verbose` flag?

lib/reporters/cli/index.js
```diff
@@ -147,7 +148,22 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
     emitter.on('beforeRequest', function (err, o) {
```
codenirvana: isn't timestamp needed for other events like the script, console, etc?

lib/reporters/cli/index.js
```diff
@@ -147,7 +148,22 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
...
         if (err || !o.request) { return; }
 
-        print('  %s %s ', colors.gray(o.request.method), colors.gray(o.request.url.toString()));
+        if (reporterOptions.showTimestamp || reporterOptions.showTimestampUtc) {
+            var currentTime = new Date();
+
+            timestamp = reporterOptions.showTimestampUtc ?
+                currentTime.toUTCString() :
+                currentTime.toLocaleString();
+            print('  %s  %s %s ',
```
codenirvana: should timestamp be wrapped in `[` brackets `]`?

lib/reporters/cli/index.js
codenirvana: I meant having multiple options/formats for logging timestamps.

README.md
jimc404: 1. I couldn't decide between choosing `timestamp` and `timestamps`. I went with the former, but since you pointed it out I guess it feels out of place for someone reading it the first time. I'll change it.


2. Since this is a CLI reporter specific option I feel `--timestamps` may be a bit confusing as other flags with `--` work on all reporters.

README.md
jimc404: Alright, I'll fix this

lib/reporters/cli/index.js
jimc404: I implemented what the PR asked for, though adding timestamp for all CLI reported events does seem to have advantages (like checking out what part of the collection is slowing things down). 

It can be added but should I add it in the current flag or will it be okay to add it in the current flag. I'd like some input on this :)

lib/reporters/cli/index.js
jimc404: Alright, I'll add brackets and see how it looks. Will commit it before the next review.

jimc404:
@codenirvana I made these changes in the latest push:
- Removed the `utc` timestamp flag
- Renamed the flag to `-show-timestamps`
- Timestamp is now shown in square brackets
Example.
```
→ https://reqres.in/api/users?page=2
  [3/23/2020, 12:39:58 AM]  GET https://reqres.in/api/users/2 [200 OK, 1.04KB, 4.3s]
```
- Added timestamps for `console` and `assertion` events
    - Timestamps could have been added for `beforeItem`, `beforeIteration` and `test` events but they by themselves completed instantly not to mention the events before and after them already had timestamps, so I skipped them. `script` event is already covered by `assertion` event so it wasn't needed. 
    - Date isn't shown for these events (only the time is). This is because the date information was becoming too redundant, date is only shown in the `beforeRequest` event

Let me know if anything else can be fixed

elit-altum:
@DannyDainton is this feature request gone stale or can I work on it?

codenirvana (approved):

PR #2251, opened 2020-03-03
Changed escape character for CSV files to double quotes

vikiCoder:
- Previously `\` was used as an escape character which is changed to `"`
- Now it allows inconsistent columns count among rows in CSV files
- Use `relax` option of `csv-parse` library to allow using double quotes `"` without escaping inside an unquoted field.

### DIFF
--- a/lib/run/options.js
+++ b/lib/run/options.js
@@ -146,12 +146,6 @@ var _ = require('lodash'),
             return parseFloat(value);
         }
 
-        if (this.cast_date) {
-            var m = Date.parse(value);
-
-            return isNaN(m) ? value : new Date(m);
-        }
-
         return value;
     },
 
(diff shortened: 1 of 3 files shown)


shamasis:
@vikiCoder can I have a bit more context of what’s happening?

1. Why are we doing this? Is it to have parity with app?
2. Will this break current newman users? How are we tackling that?
3. Why are there so many other changes in the diff beyond what’s described in PR?

vikiCoder:
1. Yes, we are doing this to have parity with the Postman app.
2. Yes, this has the potential to break current CSV files that are users using and contains `\` as an escape character. But this behaviour has disparity with the Postman app. So to fix it, maybe we can change the behaviour and add a CLI option to use old behaviour? @codenirvana 
3. Other changes are the test files updated according to the app
4. There are no GitHub issues for Newman. This change is only meant to keep parity with the app.

shamasis:
I see, `relax` related changes that I couldn’t correlate to escape character at first glance. Am I missing something when I assumed that they are “extra” changes than what’s described?

vikiCoder:
`Relax` is also related to `"`. It basically allows using `"` inside unquoted field without escaping it.

PR #2264, opened 2020-03-12
Multiple file support

### LINKED ISSUE #2215: passing multiple csv files to a collection runner in newman
How to pass multiple csv files in newman run?

newman run CollectionRunner.json --global-var IP_PORT="http://127.0.0.1:8088/" -r htmlextra --reporter-htmlextra-title "Result Report" --reporter-htmlextra-export /home/reports/HARISH.html --iteration-data /home/dataset1.csv

I want to pass multiple data sets to the collection, I have dataset1,dataset2,dataset3,dataset4 in my directory. all should gets executed for the collectionRunner and the RESULT REPORT should contain all the result,

I tried something like this, newman run CollectionRunner.json --global-var IP_PORT="http://127.0.0.1:8088/" -r htmlextra --reporter-htmlextra-title "Result Report" --reporter-htmlextra-export /home/reports/Result.html --iteration-data /home/dataset1.csv --iteration-data /home/dataset2.csv --iteration-data /home/dataset3.csv

But its not working

adrijshikhar:
fix #2215 

Add multiple file support with --iteration-data flag. Right now it checks whether the location is array or not. If its an array then check whether it has an object inside it or not. If it has no object, then it is an array of location and data is parsed for each location and then passed on, and if not an array, location variable is treated as string.

### DIFF
--- a/bin/newman.js
+++ b/bin/newman.js
@@ -26,6 +26,7 @@ program
     .option('--no-insecure-file-read', 'Prevents reading the files situated outside of the working directory')
     .option('-r, --reporters [reporters]', 'Specify the reporters to use for this run.', util.cast.csvParse, ['cli'])
     .option('-n, --iteration-count <n>', 'Define the number of iterations to run.', util.cast.integer)
+    // eslint-disable-next-line max-len
     .option('-d, --iteration-data <path>', 'Specify a data file to use for iterations (either json or csv).', util.cast.memoize, [])
     .option('--export-environment <path>', 'Exports the environment to a file after completing the run.')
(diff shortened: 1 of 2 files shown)


adrijshikhar:
the issue i am facing is that when the test is performed, in `forEach` , `async.waterfall` is asynchronous and `forEach` is synchronous. So when the test is performed, callback is called, before `forEach` getting executed and empty array is passed. I have two possible solutions for this.

1. To convert `async.waterfall` to await or promises
2. make it recursive callback function and finally return the appended values.

I prefer the recursive way. I could really use some help.
@shamasis @codenirvana

PR #2267, opened 2020-03-14
DOCUMENTATION UPDATED:

someshkoli:
External reporter filename typo. filename,csv => filename.csv


codenirvana:
Thanks for your contribution!

But, it's not a typo 😅 . Multiple reporters passed to the `-r` option are separated using `,`.

Issue #2167, opened 2019-11-06
Collection runner Log Responses


jcald1:
The Postman UI allows the user to select the following options for the collection runner in the 'Log Responses" option:  'For All Requests', 'For Failed Requests', 'For No Requests'.  We should add the same options to Newman instead of just having the '--silent Disable terminal output' option

aquibbaig:
@jcald1 I can take this issue up.

dmahajan980:
Hello everyone! I would like to work on this issue.
Thanks

jcald1:
@aquibbaig , @dmahajan980 , look like a related issue was opened: https://github.com/postmanlabs/newman/pull/2270.  Do you still want to work on this one?

PR #2270, opened 2020-03-15
Reporter for unwrapped console logging

### LINKED ISSUE #2255: (Feature request) Add cli reporter option to print plain unwrapped console logging
## TL;DR:
Let's add a `cli` reporter option that prints plain and unwrapped console logging.

## Purpose
Right now it looks like this:
```
GET http://somerequest
┌
│  <something i console.log'd out>
└
```

I’d like to have the option to just plainly print the output:
```
GET http://somerequest
<something i console.log'd out>
```

Related community discussion [here](https://community.postman.com/t/remove-from-newman-cli-output/11229).

## Proposed solution

A reporter option `cleanConsole` that can be added to as flag to the `newman run` cli command like `--reporter-cli-clean-console` and `--reporter-clean-console`.

VibhorCodecianGupta:
fixes #2255 

Read issue for description
@shamasis @codenirvana This reporter normally pipes the log output to a file that can be specified via the export option. If the user doesn't use the `clean` flag, the output is somewhat like the following screenshot, including the collection name, item name, request details, and corresponding `console` events.

Screenshot 2020-03-15 at 3 45 41 PM

If the user gives the `clean` flag, only the `console` events are recorded, like so:

Screenshot 2020-03-15 at 3 45 54 PM

I was thinking of putting in one more flag, `update`, which could allow users to write to an existing file if set to true, otherwise create a new file. Also, I wasn't sure how to get this reviewed so I just put the index.js file of the reporter in a branch and opened a PR. I have tested locally by packaging this though.

#2167 also has some interesting ideas which can be easily incorporated here. Thoughts?

### DIFF
--- a/lib/reporters/logger/index.js
+++ b/lib/reporters/logger/index.js
@@ -1,0 +1,98 @@
+var _ = require('lodash'),
+    PostmanLoggerReporter;
+
+/**
+ * Reporter that simply logs the console calls to file (default: newman-run-logs.txt).
+ *
+ * @param {Object} emitter - The collection run object, with event hooks for reporting run details.
+ * @param {Object} reporterOptions - CLI reporter options object.
+ * @param {Object} collectionRunOptions - A set of collection run options.
+ * @param {String} collectionRunOptions.export - The path to which the logs must be written.
+ * @param {Boolean=} reporterOptions.clean - Boolean flag to log only console events.
+ * @returns {*}
+ */
+
+PostmanLoggerReporter = function (emitter, reporterOptions, collectionRunOptions) {
+    var collectionName = collectionRunOptions.collection.name,
+        preRequestLog,
+        requestLog,
+        testLog,
+        runState,
+        requestMethod,
+        requestUrl,
+        itemName,
+        finalLog;
+
+    emitter.on('beforeItem', function () {
+        preRequestLog = '';
+        requestLog = '';
+        testLog = '';
+        runState = 'pre-request';
+        itemName = '';
+        finalLog = [];
+    });
+
+    emitter.on('beforePrerequest', function (err) {
+        if (err) { return; }
+        runState = 'pre-request';
+    });
+
+    emitter.on('beforeRequest', function (err, o) {
+        if (err) { return; }
+        runState = 'request';
+        requestMethod = o.request.method;
+        requestUrl = o.request.url.toString();
+    });
+
+    emitter.on('beforeTest', function (err) {
+        if (err) { return; }
+        runState = 'test';
+    });
+
+    emitter.on('item', function (err, o) {
+        if (err) { return; }
+
+        itemName = `${o.item.name}\n `;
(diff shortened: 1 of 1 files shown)


shamasis:
This is great. And I would like you to rethink the output like a product from users standpoint. What “clean” means to you, does not mean the same to someone who is using this tool. 😊

Also, think what work would user want to do and as such this will be useful for them. With that line of thinking, maybe even including timestamp is useful. It is also useful to club and group the output to do a stage change from no-log phase to log phase while running a collection. That implies that by default, logging only the timestamp, item path and event names under that path and then the logs in them is most useful as default output. 

Regarding the file management, it should create file if not exist and then append to it. One `overwrite` reporter option can be given to clean and start in a file instead of appending.

lib/reporters/logger/index.js
```diff
@@ -0,0 +1,104 @@
...
+            itemLog = itemName + PreReqLog + reqLog;
+
+        if (reporterOptions.clean) {
+            itemLog = ` ${preRequestLog}${requestLog}${testLog}`;
+        }
+
+        finalLog.push(itemLog);
+    });
+
+    emitter.on('console', function (err, o) {
```
shamasis: Log level is a very useful information for allowing to filter via options and to post that in default output.

lib/reporters/logger/index.js
```diff
@@ -0,0 +1,104 @@
...
+        if (err) { return; }
+
+        _.reduce(finalLog, function (compiledLog, log) {
+            compiledLog += `${log}\n\n`;
+
+            return compiledLog;
+        }, '');
+        finalLog = (!reporterOptions.clean) ? `${collectionName} \n\n ${finalLog}` : finalLog;
+
+        emitter.exports.push({
```
shamasis: Aside: @codenirvana ... we should build a feature where a plugin can provide a stream to put into the export. That way we do not need to store all data in memory while writing to file.

lib/reporters/logger/index.js
```diff
@@ -0,0 +1,104 @@
...
+ * @param {Boolean=} reporterOptions.clean - Boolean flag to log only console events.
+ * @returns {*}
+ */
+
+PostmanLoggerReporter = function (emitter, reporterOptions, collectionRunOptions) {
+    var collectionName = collectionRunOptions.collection.name,
+        preRequestLog,
+        requestLog,
+        testLog,
+        runState,
```
shamasis: Run state management is better handled as a enumerated structure

shamasis:
I’m also not seeing any test. Assuming they would come once the output is finessed.

shamasis:
Let’s concentrate on the base feature first and get it done perfectly. The embellishments can come after that. 

Base feature is to have a reporter where console is the hero. So, think of all uses of console in newman and then ... how users will use this using what information in the output. Remember not to step out of this main purpose because there are other reporters and places where some information is already available.

codenirvana:
Closing because of inactivity.

Also, this should be published as a separate reporter.

Issue #2244, opened 2020-02-28
[FEATURE REQUEST] Auto-suggest for newman commands 


TheDhejavu:
### This is a feature Request
What if Newman CLI could have auto-suggest for CLI commands like git commands 
E.G user could mistype help as hlp  

#### Current behavior
The current behavior currently display the below information in the screenshot to the user when a command is mistyped with a missing character 

![newman-auto](https://user-images.githubusercontent.com/19678915/75592349-5cf48b00-5a82-11ea-9909-de04f20182c5.PNG)

#### Proposed behavior using git as a case study 
![auo-newman](https://user-images.githubusercontent.com/19678915/75592480-b8bf1400-5a82-11ea-9df0-e595d53a2d3d.PNG)

gittysachin:
I would like to work on this issue.

shamasis:
Sure. But do note that newman has a very few commands to do autocomplete. There are far higher user impact features and issues that can be tackled instead😊.

If you do work on this, please do read https://github.com/postmanlabs/newman/blob/develop/.github/CONTRIBUTING.md

Post an expected behaviour and also post how someone else will maintain this feature going forward. Which terminals will this work in and what would be the limitations, if any.

gittysachin:
@shamasis Thanks. I'll post all the details about this new feature.

TheDhejavu:
@gittysachin oh well, I waited patiently for a review of this feature before suggesting an implementation but I guess you talked about implementing it first. I hope to see what you create ❤️

gittysachin:
@TheDhejavu I tried to make changes. I think that I coded correctly. But how can I build this and test on my new changes? Cause it's just working on the last version as I cloned the repo.

TheDhejavu:
@gittysachin correct me if I'm wrong, you mean you have been working on a few changes before this and now you trying to implement this new feature on top of these changes? Or you are just starting out and trying to figure out how to build and test it on the new changes made to the repo ?

gittysachin:
@TheDhejavu I tried changing the code and I think I wrote it correctly. But now it's not doing anything new when I'm trying `node ./bin/newman.js v`. So is there any way I can try with my new changes? It's just reacting as the old.

TheDhejavu:
Oh, i get it now. try using a different command like `node ./bin/newman.js --hlp` , it should give a list of defined suggestion like
`do you mean help `

gittysachin:
[...]

PR #2271, opened 2020-03-15
Feature: Auto Suggest for Newman Commands

gittysachin:
Fixes [#2244](https://github.com/postmanlabs/newman/issues/2244)

**What does this PR do?**
I have implemented the command auto-suggest feature for Newman.

**Implementation:** 
Now if someone uses wrong option in command for Newman CLI then it will suggest similar commands if there are any.

**Testing:**
We have `--help` option for newman CLI command. So if we use wrong option as command like `node ./bin/newman.js --hlp` then it will suggest an option `--help` as most similar command.


codenirvana:
Closing as per https://github.com/postmanlabs/newman/issues/2244#issuecomment-683410882

Issue #2262, opened 2020-03-11
Run multiple collections sequentially sharing a common environment.


lusoalex:
Dear,

I was wondering if it is possible (did not find how) to run sequentially two or more collections, but sharing the same environment.

Something like 
```bash
newman run authenticate_collection.json api_collection.json -e environment.json
```

I have a collection that basically login a user and retrieve an access_token.
Then I have multiple collection for several APIs.
I would not to have to copy/paste each time the authentication folder but share it instead.
So the result (access_token) of the authenticate_collection.json would be into the environment, only for the run execution context.

I guess it require to be handle first by postman before newman?
Is this something possible?
Thanks a lot for your help!

shubhamjadon:
@lusoalex  i do not think that running multiple collections sequentially is possible in newman. This feature needs to be added to newman.(something similar was requested in issue #2117 ).

lusoalex:
Hello, thanks for your quick answer. I really appreciate!

It is not totally similar to issue [#2117 ](https://github.com/postmanlabs/newman/issues/2117) as it is not just run collection one by one in a loop, but keep track of new values into the environment (only during the newman execution lifetime)

Regarding the example I gave on the issue, the authentication_collection.json will add an access_token into the environment, that would be used by the api_collection.json.
But in the end, I do not expect the environment to keep this value once newman has finished.

To be transparent, this is because I am developing +10 different APIs and It's annoying to copy/paste the folder containing the request to obtain the access_token. Plus when there's an update on these requests, I have to apply them everywhere.

Please could you tell me if this feature seems "strange/useless" or if you have some workaround?
By the way, I guess it should be implemented first in postman before newman right?

Thanks a lot for you time and expertise!
Alexandre.

shubhamjadon:
@lusoalex there is no feature like the one you asked currently in postman.
But there is a workaround i discovered maybe that could help you. It is as follows:
1. You can change environment variable using Test Scripts(go through this [link](https://learning.postman.com/docs/postman/collection-runs/using-environments-in-collection-runs/) for tutorial). The environment variable will get changed after getting a response from collection run.
2. Then you can use that value in other collection run. As you mentioned it is a access token so you can either add it to url or header of requests.([more info](https://www.youtube.com/watch?time_continue=432&v=HiIJbmjeTPU&feature=emb_title))
[...]

PR #2273, opened 2020-03-16
Feature/multiple collection run mentioned in issue #2117

### LINKED ISSUE #2117: Feature request: run with multiple files
Hello,
the `run` command could accept as input a regexp or a list of file in order to execute all the collection as a single one.

_Use case_

If you have a huge test suite you need to break up into multiple files.
Right now I have 140k request in 14 `.postman_collection` files.

You could use:

`for fn in `ls *.postman_collection`; do newman run $fn --env-var 'baseUrl=http://localhost:3000'; done`

But the bail logic is gone and I need bash scripting to manage the exit code and reports.

WDYT?

shubhamjadon:
Fixes #2117 
It implements feature requested in issue #2117 . After merging following changes developers will be able to run multiple collections as single one by just passing list of collection objects in Newman library.
Like this:
```
newman.run({
    collection: [require('./test1.json'),require('./test.json'),require('./test2.json')],
    reporters: 'cli',
    iterationCount: 2,

}, function (err) {
	if (err) { throw err; }
    console.log('collection run complete!');
});
```


shubhamjadon:
@codenirvana, sure i will start working on some other approach to solve this issue.

And i thought of merging all the collections into one because of the following reasons:

- As in the issue #2262 , the user also wanted to run multiple collection together but he also needed the ability to access environment variables of one collection in other for the the purpose of authentication . So, to make that functionality available i thought it would be better to merge all collections into a single one.
- As you mentioned that this approach will be very confusing, to make it less confusing i made sure that all the collections are represented like separate folders inside the new collection created. So, user can make sense of which request belongs to which collection.
- And as you mentioned that such an interface can break a lot of existing reporters. So, i think it should not break any reporters as i am overwriting the options.collection with a combined collection in the very starting of the code so that it does not affects the later code and later code considers combined collection as the input collection.

Kindly share your review on following.

shubhamjadon:
@codenirvana As suggested by you I found a new approach to solve issue #2117 it is as follows:
```
const newman = require('newman');
let collections = ['./test.json','./mycollection.postman_collection.json'];

(function runCollection(i = 0) {
    newman.run({
        collection: collections[i],
        reporters: 'cli',
        iterationCount: 1,
    }, function (err) {
        if(err){ throw err; }
        i++;
        if(i<collections.length){
            runCollection(i);
        }
        else{
            console.log("All Collections Run Complete!")
        }
    });
})();
```
This code is suitable for both single collection and multiple collections run. So, we can update the readme according to above.
PTAL and notify me , I will start pull request with the updated ReadMe.

cupcakepanda89:
@shubhamjadon

Do you know if it is possible to pass the url-with-collection-id to this line `collection: collections[i]` ?

So I don't have to export the JSON files everytime there are changes.

shubhamjadon:
@cupcakepanda89 
you can give a string of the URL where the Collection JSON can be found or path to a local JSON file.

abhisheksingh999:
Hello @shubhamjadon thank for this. It is helpful. 
how do I run the above JavaScript from Windows command prompt?

abhisheksingh999:
I get this error when i run npm run myLibrary.js
image

abhisheksingh999:
@shubhamjadon How can I disable SSL, while running these collections. I used to use --insecure running single collection.

shamasis:
Closing since the PR has the README change as well as code change. 

From what I can read at a glance, the original approach to merge collections isn't an advisable solution. The alternate is to provide simple example in wiki to have a script to use Newman programmatically. 

PS - A lot of core assumptions have to change to make Newman run multiple collections by itself. As such, that doesn't seem something that will be built in reasonable future.

vijayarondla-mosaic:
@shubhamjadon  HI I was wondering how can we pass the Environment json file as an input . I am able to pass collections as url but i need to pass env also .If i give at the end of the collection string, it errors out . Thanks

vijayarondla-mosaic:
I think i saw the documention let me try

Issue #1599, opened 2018-06-28
Postman - Newman Error:407 Proxy Authentication Required


SMES03:
Hi everyone,

Informations first:

 1. Newman Version (can be found via `newman -v`):3.9.4
 2. OS details (type, version, and architecture):Windows Server 2016 x64
 3. Are you using Newman as a library, or via the CLI? CLI
 3. Did you encounter this recently, or has this bug always been there: this bug always been there (but i use it only since few days)
 4. Expected behaviour: a response from the server (with status 200)
 5. Command / script used to run Newman: something like

`C:\Users\[USER]\AppData\Roaming\npm\node_modules\newman\bin\newman.js run C:\collection\AuthentificationTFS.postman_collection.json -e C:\environment\DevCI.postman_environment.json -g 'C:\global\My Workspace.postman_globals.json'`

for each request i have the same result (Request POST and/or GET)

here is the result
![capture1](https://user-images.githubusercontent.com/40665114/42037969-fba3e072-7ae9-11e8-8109-44c1dc27b6a3.JPG)

The same request send by browser works fine
![capture2](https://user-images.githubusercontent.com/40665114/42037979-0556514a-7aea-11e8-8cb7-0872e7d20a46.JPG)

For your information, the server who receive the requests is in local network.

Thanks

kunagpal:
Hey @SMES03, it looks like you're using Newman within a Proxy network environment, and that your proxy requires authentication. Unfortunately, Newman does not support Proxy authentication at the moment. I'm more than happy to mark this as a feature request, though 😄

codenirvana:
@SMES03 As a workaround, check if setting `HTTP_PROXY`/`HTTPS_PROXY` environment variable fixes this issue?

```
HTTP_PROXY=http://USERNAME:PASSWORD@PROXYIP:PROXYPORT
```

doughless:
I tested the workaround above, and it does work with our corporate proxy. That said, it would be preferable to not have to store passwords in environment variables.

someshkoli:
I

someshkoli:
I just got started to add proxy auth feature to newman cli.

someshkoli:
Added a proxy parameter to cli optioins as shown below:
`node bin/newman.js run examples/sample-collection.json --proxy username:password
`
is this what was required ?

codenirvana:
Closing this issue as we are not able to reproduce this internally and we haven't seen other users facing the same.

[...]

PR #2276, opened 2020-03-16
Feature/proxy authentication for newman CLI 

someshkoli:
Fixes Issue / Feature request : #1599

As the issue stated there was no other way to use newman  CLI in a proxy environment rather than setting proxy authentication credentials in the environment. This is not so secure in some cases. To avoid this one way is to supply proxy credentials via CLI as an option to the newman. 

This PR provides this solution by supplying proxy authentication credentials via CLI as show below
`newman run sample-collection.json --proxy username:password`

Running collection without proxy set:

withouproxy

Running collection with proxy set:
with proxy


README.md
```diff
@@ -494,7 +494,7 @@ Several members of the Postman community have created custom reporters offering
 Once the custom reporter NPM package has been installed either globally or locally, this can be then used with Newman in the following ways:
 
 ```console
-$ newman run /path/to/collection.json -r htmlextra,csv
+$ newman run /path/to/collection.json -r htmlextra.csv
```
shamasis: This isn’t an error. -r htmlextra,csv implies using two reporters.

shamasis:
@codenirvana @codeitva - can you point to the right runtime proxy input code / doc?

README.md
someshkoli: Oh damn. I totally miss  read that. My bad

someshkoli:
Okay ill check for the runtime options and push the proxy to them directly. Also provide option to specity proxy authentication method

someshkoli:
@shamasis  What all proxy authentication methods should be added?  Also in the sdk(postman-collection)  there is no way to specify the authentication method. so the default which postman is using right now is basic.

shamasis:
These are options passed to runtime when a new run instance is created.

someshkoli:
> These are options passed to runtime when a new run instance is created.

Nah i was talking about how the user will supply these values to newman

shamasis:
That’s the whole reason we didn’t reinvent the wheel by adding in CLI and instead left it as all other CLI tools do ... using environment vars!

I wouldn’t over complicate this feature. I would either put clean CLI options or not build any CLI options. 😛 JSON source will be an overkill.

someshkoli:
xD okay then CLI options it is. Also could you tag someone who could help with specifying auth. I'm having trouble in pin pointing where to add the auth options.Thanks

someshkoli:
@shamasis can you please review this ?

codenirvana (commented):
@someshkoli Great work but we are not planning to support this option at the moment.

Currently, one can configure proxy using `HTTP_PROXY`/`HTTPS_PROXY` environment variables and we haven't seen a lot of users requesting this so closing this for now.

swatig90:
Hi @someshkoli , I am trying to execute some requests which are running behind proxy and facing exactly same issue which is described here and also issue 1599.
Could you please let me know whether sending proxy along with newman run command is possible now, or if there is any other alternative.

I already tried --env-var to set HTTP_PROXY/ HTTPS_PROXY, but it dint help.

Error message I receive is: tunneling socket could not be established, statusCode=403

Many thanks in advance!

someshkoli:
> Hi @someshkoli , I am trying to execute some requests which are running behind proxy and facing exactly same issue which is described here and also issue 1599.
> Could you please let me know whether sending proxy along with newman run command is possible now, or if there is any other alternative.
> 
> I already tried --env-var to set HTTP_PROXY/ HTTPS_PROXY, but it dint help.
> 
> Error message I receive is: tunneling socket could not be established, statusCode=403
> 
> Many thanks in advance!

Yeah you can do that, but you need to pass that var to your shell's env not as a flag
You can easily do this by temporarily passing them using `export` on linux shell or permanently storing them in your shell's rc file and equv in windows too

vbisbest:
I really wish this would be accepted.  Using environment variables for this can have consequences.  What about after the run and removing the variables?  Chrome uses the same environment variables and you may not want chrome to use a proxy.   CLI option is the cleanest way to do this especially when automating newman.

swatig90:
Suggested solution by Someshkoli had worked well for me. This is how I implemented. I setup HTTP_PROXY and HTTPS_PROXY environment variable.
Another alternative is setup proxy variables at npm level.

Chrome will not be impacted via any of these (it dint at least in my case), however access to other APIs which do not require proxy, might be impacted.

PR #2277, opened 2020-03-17
Feature: skip responses option for the JSON reporter

### LINKED ISSUE #1510: JSON output is huge (120mb) vs JUnit (24kb)
1. Newman Version (can be found via `newman -v`): **3.9.3**
 2. OS details (type, version, and architecture): **OSX Sierra**
 3. Are you using Newman as a library, or via the CLI? **CLI**
 3. Did you encounter this recently, or has this bug always been there: **n/a**
 4. Expected behaviour: **much smaller JSON reporter output**
 5. Command / script used to run Newman: `newman run fapi.postman_collection.json -e cc-production.postman_environment.json --reporters cli,json --reporter-json-export results.json`

We're trying to do automated API Response time tests via newman and are re-using a postman_collection that we use for testing. The plan is to take the JSON output, parse it and output the response times somewhere.

Some API calls have fairly large responses like 1mb and it seems the JSON reporter outputs a Stream Buffer of the response which is absolutely huge.

- JSON: **120mb** `newman run fapi.postman_collection.json -e cc-production.postman_environment.json --reporters cli,json --reporter-json-export results.json`
- JUnit: **24kb** `newman run fapi.postman_collection.json -e cc-production.postman_environment.json --reporters cli,junit --reporter-junit-export results.xml`

I suppose the biggest WTF was that the JSON Reporter includes Stream buffer responses for each API call

jimc404:
Fixes #1510
### What does this PR do?
This implements the `--reporter-json-skip-responses` in the JSON reporter (and consequently `skipResponses` key when using the library). The primary aim is to reduce the huge files size that JSON reporter causes by dumping the response body into the JSON report.

### Implementation
In the JSON reporter, if the related key was present then a deepClone of the run summary is made with lodash and response body is removed from that object. The resulting object was sent to be written into the exported file.

### Testing
As there wasn't much testing being done for the JSON reporter I created 2 filed in the `test` directory namely `test/cli/json-reporter-response.test.js` and `test/library/json-reporter-responses.test.js`. They test the JSON reporter with and without the skip responses option passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -442,6 +442,7 @@ To enable JSON reporter, provide `--reporters json` as a CLI option.
 | CLI Option  | Description       |
 |-------------|-------------------|
 | `--reporter-json-export <path>` | Specify a path where the output JSON file will be written to disk. If not specified, the file will be written to `newman/` in the current working directory. If the specified path does not exist, it will be created. However, if the specified path is a pre-existing directory, the report will be generated in that directory. |
+| `--reporter-json-skip-responses` | Skip API responses when writing output to file | 
 
 ### JUNIT/XML Reporter
 The built-in JUnit reporter can output a summary of the collection run to a JUnit compatible XML file. To enable the JUNIT reporter, provide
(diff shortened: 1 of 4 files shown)


prakashn27:
May I know which package this change will be available in?

PR #2278, opened 2020-03-18
Feature/script details: Allows user to see the origin of each script

### LINKED ISSUE #2022: Newman CLI requires 200 status code to pass test
I started by writing the failing test case, checking if my endpoint 401s at the right time. Despite not adding a 200 check, newman seems to be automatically inserting that into every test.

```
↳ Fail to get Client AT in Authorization Body w/ Bad Key and Secret
  POST http://localhost:5001/api/v1/accessToken [401 Unauthorized, 350B, 16ms]
  6. Status code is 200 // THIS IS THE FAILING TEST
  ✓  Body should be Unauthorized
  ✓  Status code is 401
```

 1. Newman Version (can be found via `newman -v`):
 3. Are you using Newman as a library, or via the CLI?
 4. Expected behaviour:

narayanpai1:
Fixes #2022 

### What does this PR do?

1. It lets the user see the location of each script in the line following the assertion or script-error when the verbose option is true.

2. It hints at the location of each script by a symbol corresponding to it.
The symbols used are:
   -  `symbols.root` for a script in the collection.
   -  `symbols.folder` for a script in a folder.
   -  space(blank) for a script in the item itself.

### Screenshots

1.  With verbose false
   - Before
before

   - After
after

2.  With verbose true
   - Before
before_with_verbose

   - After
after_with_verbose

### Implementation

1.  Added a function in cli-utils to get the location and the symbol. It takes in `PostmanItem` on which the script is executed, `scriptId` and the `symbols` object with the required encoding.
2. Use the return object before and after each assertion and script-error.

### Testing

For testing this function, a unit-test has been added in a new file `test/unit/cli-utils.test.js`. It tests the function for all three levels of scripts.

### DIFF
--- a/lib/reporters/cli/cli-utils.js
+++ b/lib/reporters/cli/cli-utils.js
@@ -1,4 +1,5 @@
-var inspect = require('util').inspect,
+var util = require('../../util'),
+    inspect = require('util').inspect,
     wrap = require('word-wrap'),
     symbols = require('./cli-utils-symbols'),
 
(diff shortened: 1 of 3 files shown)


narayanpai1:
The tests are failing in Linux environments since the coverage in functions in test-library does not meet the global threshold. Since my changes are only applicable to cli, I don't know how to fix this. Can you please suggest me how to proceed @shamasis

shamasis (changes requested):
It seems like the implementation can be slightly improved by making Assertion event itself send this data instead of finding it in a roundabout fashion it is needed now.

PR #2349, opened 2020-05-13
Allow newman run to accept requestAgent option

gone-skiing:
Companion PR to https://github.com/postmanlabs/postman-runtime/pull/1014 (included in postman-runtime version 7.25.0) to address the need to run postman tests via socks proxy. It is a long-standing request documented here: postmanlabs/postman-app-support#1321.

The proposed first (minimum viable) step is to allow teams that use newman as a library to pass their own HTTP agent through to postman request. Postman request already supports agent option. This would enable use of libraries like socks-proxy-agent to run tests through socks proxy.


gone-skiing:
@codenirvana now that postman-runtime is released and included in newman's package.json, could we push this PR through?

lib/run/index.js
```diff
@@ -154,6 +154,7 @@ module.exports = function (options, callback) {
 
         // now start the run!
         runner.run(options.collection, {
+            agent: options.agent,
```
codenirvana, line 157: It's better to expose Newman option as `requestAgent`

codenirvana:
@gone-skiing would you like to integrate support for the proxy agent within Newman?

gone-skiing:
@codenirvana that is our use case. Use `newman` as library to run collections.

gone-skiing:
@codenirvana thanks for the feedback, that is a better option name for sure. Updated.

gone-skiing:
@codenirvana just checking to see if you need anything on this PR... Please let me know. Cheers!

codenirvana:
@gone-skiing Yes, I was trying out this change and it appears to me that we need to add support for multiple agents in the postman-runtime library otherwise it will only server either HTTP or HTTPS requests.
I am working on this change right now, will let you know if there's any update.

gone-skiing:
@codenirvana checking in on this. We are blocked on our end unless we have a way to pass in at least one agent. Anything we can do to help move this forward? Could we merge this as first step to unblock downstream development?

Our understanding was that if an agent is passed in, it should take care of all request types. It is certainly the case for our scenario. 

To make sure we understand your concern, are you worried about cases where there are http/https specific settings that require their own agents and API consumer also specifies agent option? 

Thanks for all your efforts to keep this moving. Much appreciated!

codenirvana:
@gone-skiing Thanks for keeping the patience, the ability to accept custom requesting agents have been added to Newman v5.1.0.
Here's an example of setting up Socks Proxy: https://github.com/postmanlabs/newman#using-socks-proxy

PR #2373, opened 2020-05-27
Modify code to use ES6 features

narayanpai1:
### What does this PR do?
It modifies some parts of the codebase to use new ES6 features like class, arrow functions, async-await etc.


narayanpai1:
@codenirvana Can I start replacing functions with callbacks to the async-await styled ones.
To do so, I would have to add the following rule in eslintrc file:
```
{
    "parserOptions": {
        "ecmaVersion": 2018
    }
}
```

PR #2395, opened 2020-06-09
Add RC file controller

narayanpai1:
### What does this PR do?
It adds functions for loading and storing configurations from different RC files across the system.

### Minor details
1. The function for loading returns an empty object if the file doesn't exist.
2. The function for storing creates the file if it doesn't exist. 
   - In *nix based systems, the new file created has read and write permissions only to the owner of the file.
   - In Windows, since the file is stored in the User directory, it cannot be read by any other user.

### Testing
1. The unit tests for both the functions are added.
   - The whole testing process is done by changing the environment variable providing the home directory, to the `outDir` used in other tests.

### DIFF
--- a/lib/config/rc-file.js
+++ b/lib/config/rc-file.js
@@ -1,59 +1,159 @@
 /* eslint-disable no-process-env */
 var _ = require('lodash'),
     fs = require('fs'),
+    waterfall = require('async/waterfall'),
     join = require('path').join,
     async = require('async'),
     util = require('../util'),
     liquidJSON = require('liquid-json'),
+    os = require('os'),
 
     /**
-     * Name of the directory that contains the file denoted by FILE_NAME.
+     * Name of the directory that contains Postman config files.
      *
      * @type {String}
      */
-    POSTMAN_CONFIG_DIR = 'postman',
+    DIR_NAME = 'postman',
 
     /**
-     * Name of the file that contains Newman compliant confguration information.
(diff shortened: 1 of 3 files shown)


narayanpai1:
@codenirvana I will address the issues raised in the PR for login command after this PR is merged. Is that fine?

lib/config/rc-file.js
```diff
@@ -1,59 +1,143 @@
...
+     * Configuration loader to acquire and merge settings from the requested config files.
+     *
+     * Data corresponding to an inexistent file is taken as an empty object
+     *
+     * @param {String[]} [types=['home']] - Types of the config files to be read.
+     * @param {Function} callback - The callback function invoked to mark the completion of the config loading routine
+     * @returns {*}
+     */
+    load: (types, callback) => {
+        (!callback && _.isFunction(types)) && (
```
codenirvana, line 63: if block will be better.

lib/config/rc-file.js
```diff
@@ -1,59 +1,143 @@
...
+        );
+
+        let file = _.get(configFiles, type);
 
-        return callback(null, _.merge.apply(this, files));
-    });
+        waterfall([
+            (next) => {
+                // ensure the config directory exists if `type` is home
+                return type === 'home' && !fs.existsSync(HOME_CONFIG_DIR) ?
```
codenirvana, line 119: Simplify this using if blocks.

PR #2401, opened 2020-06-13
Add logout command

narayanpai1:
### What does this PR do?
It adds the logout command. The command lists out all the alias available in the RC File located in the home directory and prompts the user to select an alias to delete.

### Usage
```console
newman logout
```

### Screenshots
image

### Testing
- The CLI and parser tests for the command are added.

### DIFF
--- a/.nycrc.js
+++ b/.nycrc.js
@@ -19,7 +19,7 @@ function configOverrides (testType) {
                 branches: 20,
                 functions: 35,
                 lines: 35,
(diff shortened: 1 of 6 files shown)


narayanpai1:
@codenirvana Since the only additional code being checked in library coverage tests is in `bin/newman.js`, I don't feel the coverage for `test/library` can be increased. Hence, I updated the coverage threshold.

PR #2416, opened 2020-06-29
Add api-key-alias option

narayanpai1:
### What does this PR do?
It adds `postman-api-key-alias` option for `run` command along with its corresponding `environment` variable named `POSTMAN_API_KEY_ALIAS`. It is used to refer to a stored API Key by its alias.

Additionally, it also adds an `environment` variable named `POSTMAN_API_KEY` corresponding to the existing `postman-api-key` option.

### Minor details

1. The alias is only used when its absolutely required for authentication details.
For eg, the alias is not used in the following circumstances:
a. The API-Key is directly available through the option, env-variable or as a query parameter in the URL.
b. The collection referred to doesn't refer to a remote collection.
2. When the API-Key stored is encrypted, the user is prompted for a passkey to decrypt it.
3. When none of the aliases are referred and there is no API-Key available, the program assumes `default` alias.

### Packages added
An npm-package named [prompts](https://www.npmjs.com/package/prompts) is added as a dependency to prompt for the passkey.
The prompts in the login and logout command will also be replaced using the same library in the following PR.

### Screenshots
1. Using the option
image

2. Using the `ENV` variable
Note the use of a passkey since the corresponding API-Key is encrypted
image

### Testing
The following tests are added
- CLI, CLI-parser tests for postman-api-key-alias option
- The CLI parser tests for the option and the use of environment variables are added
- An existing library test for postmanApiKey is edited to consider a corner case

### DIFF
--- a/npm/test-cli.js
+++ b/npm/test-cli.js
@@ -2,7 +2,8 @@
 require('shelljs/global');
 require('colors');
 
-var Mocha = require('mocha'),
+var _ = require('lodash'),
+    Mocha = require('mocha'),
     expect = require('chai').expect,
     join = require('path').join,
     recursive = require('recursive-readdir'),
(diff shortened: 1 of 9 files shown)


npm/test-cli.js
```diff
@@ -25,6 +25,11 @@ module.exports = function (exit) {
     // eslint-disable-next-line no-process-env
     process.env[isWin ? 'userprofile' : 'HOME'] = outDir;
 
+    // Used to indicate the child processes that their execution is during a test
+    // This slightly alters their logic to test some cases
+    // eslint-disable-next-line no-process-env
+    process.env.NEWMAN_TEST_ENV = 'cli';
```
narayanpai1, line 31: This is done here since `NEWMAN_TEST_ENV=cli node test-cli.js` does not work in Windows

PR #2433, opened 2020-07-10
Fix: Alter logic to load config only for CLI-run

narayanpai1:
### What does this PR do?
Currently, details from RC-file and environment are loaded even for library runs.
This PR separates out code related to config sources from the library, to not let this happen.

Additionally, the PR makes sure that options mentioned in the RC File get higher priority than the default-options.

### Implementation details
1. The `lib/config` module initially used in `lib/run/options.js` is now used in `bin/newman.js`.
2. An additional file `lib/config/defaults.js` is created to mention all the default options to give these options lesser priority than the ones in rc-file.

### Testing
1. CLI-Parser tests are added to make sure proper merging of options take place from different sources.
2. The home directory is changed to `outDir` in CLI, unit test-suites to make sure the RC file used by the developers doesn't interfere in the tests.

### DIFF
--- a/lib/run/options.js
+++ b/lib/run/options.js
@@ -7,7 +7,6 @@ var _ = require('lodash'),
     liquidJSON = require('liquid-json'),
     parseCsv = require('csv-parse'),
     util = require('../util'),
-    config = require('../config'),
 
     /**
      * The message displayed when the specified collection file can't be loaded.
@@ -338,7 +337,7 @@ var _ = require('lodash'),
     };
 
 /**
- * The helper function to load all file based information for the current collection run.
+ * The helper function to load all file/cloud-API based information for the current collection run.
  *
  * @param {Object} options - The set of generic collection run options.
  * @param {Function} callback - The function called to mark the completion of the configuration load routine.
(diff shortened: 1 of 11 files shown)


narayanpai1:
@codenirvana This PR is ready for review

lib/run/options.js
```diff
@@ -354,13 +353,27 @@ module.exports = function (options, callback) {
...
     options.insecureFileRead = Boolean(_.get(options, 'insecureFileRead', true));
 
-    config.get(options, { loaders: configLoaders, command: 'run' }, function (err, result) {
+    // sanitize environment and globals
+    !options.environment && (options.environment = {});
+    !options.globals && (options.globals = {});
+
+    let commonOptions = _.pick(options, ['postmanApiKey']);
+
+    async.mapValuesSeries(options, (value, name, cb) => {
```
codenirvana, line 362: why in series?

narayanpai1: Because the logic contains prompting for a passkey.
So once the postmanApiKey is decrypted for collection, it should not be asked for environment.

narayanpai1: So is it okay or should I find an alternate way?

narayanpai1:
@codenirvana I have performed the requested changes

PR #2447, opened 2020-07-28
Add API module to interact with Postman using APIs

narayanpai1:
### What does this PR do?
This PR adds API module to perform different operations using Postman APIs. It consists of a class representing a resource in Postman-Cloud like a collection or an environment. All the operations on a given Postman resource can be then done using member functions of the class like `get()`, `update()` etc.

### Implementation Details
- The module is integrated with `run` to fetch Postman collections and environments. 
- Once initialized before loading run-options, fetching of these resources can be done using the `get()` member function of the corresponding class instance.
- Each class instance also stores authentication details like postman-api-key. This apikey is either the apikey from the query-param of the URL or the session-api-key.

### Testing
- Unit tests for the class are added.
- Library tests on postman-api-key option are updated to address the changes.

### DIFF
--- a/lib/api/index.js
+++ b/lib/api/index.js
@@ -1,0 +1,4 @@
+module.exports = {
+    // @todo - Add PostmanAPIManager
+    PostmanResource: require('./postman-resource')
+};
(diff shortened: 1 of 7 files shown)


narayanpai1:
@codenirvana The PR is updated to address the issues raised.

PR #2486, opened 2020-08-26
Improve login user experience

narayanpai1:
### What does this PR do?
This PR improves the user experience in dealing with stored API Keys. 

### Major Changes
1. The login command now assumes that the alias is `default` if there is no data regarding it already stored. This way the user doesn't have to bother about the internal workings of the command.

2. If the `default` API Key already exists, the command will allow the user to either override it or create a new alias.

3. The logout command now gets user confirmation before deleting the API Key.

### Minor Changes
1. Better messages and prompts.
2. Improvement in the coloring of prompt messages.
3. Validation of the input for API Key

### Testing
The CLI tests of the `login` command, `logout` command and `postman-api-key-alias` commands are updated.

### DIFF
--- a/.nycrc.js
+++ b/.nycrc.js
@@ -8,10 +8,10 @@ function configOverrides (testType) {
     switch (testType) {
         case 'cli':
             return {
-                statements: 80,
-                branches: 65,
-                functions: 85,
-                lines: 80
+                statements: 82,
+                branches: 67,
+                functions: 86,
+                lines: 83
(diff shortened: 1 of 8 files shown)


codenirvana (approved):

DannyDainton:
@narayanpai1 / @codenirvana

This is a cool change 🏆 - Has the README.md been updated to explain how users can take advantage of the feature?

PR #2509, opened 2020-09-16
Fix do not store memory heavy objects in summary - instead give options

lucifercr07:
We are introducing a new flag `--skip-execution-response` on enabling which the summary response
will not be included in the report.

Also when `cli` is introduced as only reporter in the list of reporters we won't be saving execution responses.
This fix will allow developers using newman to generate the report even for large memory objects, in this case
only the summary response won't be part of the report.

Tested for reporters: `cli, html, json, junit, confluence, csv`

### DIFF
--- a/bin/newman.js
+++ b/bin/newman.js
@@ -60,6 +60,7 @@ program
     .option('--ssl-extra-ca-certs <path>', 'Specify additionally trusted CA certificates (PEM)')
     .option('--cookie-jar <path>', 'Specify the path to a custom cookie jar (serialized tough-cookie JSON) ')
     .option('--export-cookie-jar <path>', 'Exports the cookie jar to a file after completing the run')
+    .option('--skip-execution-response', 'Skips addition of response to the execution report generated')
     .option('--verbose', 'Show detailed information of collection run and each request sent')
     .action((collection, command) => {
         let options = util.commanderToObject(command),
--- a/lib/run/summary.js
+++ b/lib/run/summary.js
@@ -39,6 +39,16 @@ RunSummary = function RunSummary (emitter, options) {
         globals: _.get(options, 'globals'),
 
         /**
+         * Holds information related to the reporters specified.
+         */
+        reporters: _.isString(options.reporters) ? [options.reporters] : options.reporters,
+
+        /**
+         * Holds information related to skip addition of response to the execution report generated.
+         */
+        skipExecutionResponse: _.get(options, 'skipExecutionResponse'),
+
+        /**
          * Holds information related to the run.
          */
         run: {
@@ -223,14 +233,22 @@ _.assign(RunSummary, {
         emitter.on('request', function (err, o) {
             if (!_.get(o, 'cursor.ref')) { return; }
 
-            var execution = cache[o.cursor.ref] = (cache[o.cursor.ref] || {});
+            var execution = cache[o.cursor.ref] = (cache[o.cursor.ref] || {}),
+                executionResponse = {
+                    cursor: o.cursor,
+                    request: o.request,
+                    response: o.response,
+                    id: _.get(o, 'item.id')
+                };
 
-            executions.push(_.assignIn(execution, {
-                cursor: o.cursor,
-                request: o.request,
-                response: o.response,
-                id: _.get(o, 'item.id')
-            }, err && {
+            // check if skipExecutionResponse set in options or if cli is the only reporter specified, we do
+            // not include executionResponse as part of executions summary
+            if (summary.skipExecutionResponse || (_.isArray(summary.reporters) &&
+                summary.reporters.length === 1 && (summary.reporters).includes('cli'))) {
+                executionResponse.response.stream = Buffer.from('');
+            }
+
+            executions.push(_.assignIn(execution, executionResponse, err && {
                 requestError: err || undefined
             }));
         });


lib/run/summary.js
```diff
@@ -223,14 +233,22 @@ _.assign(RunSummary, {
...
+                    cursor: o.cursor,
+                    request: o.request,
+                    response: o.response,
+                    id: _.get(o, 'item.id')
+                };
+
+            // check if skipExecutionResponse set in options or if cli is the only reporter specified, we do
+            // not include executionResponse as part of executions summary
+            if (summary.skipExecutionResponse || (_.isArray(summary.reporters) &&
+                summary.reporters.length === 1 && (summary.reporters).includes('cli'))) {
```
coditva, line 247: We can put the value of whether the only reporter is `cli` inside a variable for readability:
```suggestion
			var hasOnlyCliReporter = _.isArray(summary.reporters) &&
		    	summary.reporters.length === 1 &&
		    	summary.reporters.includes('cli');
    	
            if (summary.skipExecutionResponse || hasOnlyCliReporter) {
```

Maybe give a little more thought to the variable name `hasOnlyCliReporter`. 😅

coditva, line 247: I think this would be faster:
```suggestion
                summary.reporters.length === 1 && summary.reporters[0] === 'cli')) {
```

Can we benchmark these two approaches? See https://benchmarkjs.com

CHANGELOG.yaml
```diff
@@ -1,3 +1,7 @@
+master:
+    fixed bugs:
+     - GH-2509 Fix do not store memory heavy objects in summary - instead give options
```
coditva, line 3: This should be an english sentence. 😄

lib/run/summary.js
```diff
@@ -38,6 +38,16 @@ RunSummary = function RunSummary (emitter, options) {
          */
         globals: _.get(options, 'globals'),
 
+        /**
+         * Holds information related to the reporters specified.
+         */
+        reporters: _.get(options, 'reporters', []),
```
coditva, line 44: Should/would this always be an array? Even if one reporter is specified? Can we add a test for this as well?

lib/run/summary.js
```diff
@@ -223,14 +233,22 @@ _.assign(RunSummary, {
...
-                cursor: o.cursor,
-                request: o.request,
-                response: o.response,
-                id: _.get(o, 'item.id')
-            }, err && {
+            var execution = cache[o.cursor.ref] = (cache[o.cursor.ref] || {}),
+                executionResponse = {
+                    cursor: o.cursor,
+                    request: o.request,
+                    response: o.response,
```
coditva, line 240: Why don't we add response only when the below `if` condition is false, instead of adding response and then changing it to empty buffer?

lib/run/summary.js
```diff
@@ -38,6 +38,16 @@ RunSummary = function RunSummary (emitter, options) {
...
+        /**
+         * Holds information related to the reporters specified.
+         */
+        reporters: _.get(options, 'reporters', []),
+
+        /**
+         * Holds information related to skip addition of response to the execution report generated.
+         */
+        skipExecutionResponse: _.get(options, 'skipExecutionResponse'),
+
```
coditva, line 50: Do we need to add the `reporters` and `skipExecutionResponse` options to the `RunSummary` object? This would change the API, no? Can we not just use them wherever required ad-hoc?

test/unit/run-summary.test.js
```diff
@@ -33,7 +33,8 @@ describe('run summary', function () {
     it('should have the relevant top-level data structures', function () {
         var summary = new Summary(new EventEmitter());
 
-        expect(_.keys(summary).sort()).to.eql(['collection', 'environment', 'globals', 'run'].sort());
+        expect(_.keys(summary).sort()).to.eql(['collection', 'environment', 'globals', 'reporters',
+            'run', 'skipExecutionResponse'].sort());
```
coditva, line 37: I don't think we should be changing the API here. This test should not need to be changed.

coditva (changes requested):
1. We should avoid changing any API here.
2. Need to have some more tests.

coditva:
@lucifercr07 this improvement works well in 2 scenarios:
1. when the only reporter is CLI.
2. when reporter is not CLI, and you don't care about the response.

Have you discussed this solutions with @codenirvana? Did we explore other solutions like, maybe storing response in secondary memory and reading them at the end when generating the summary?

lib/run/summary.js
```diff
@@ -12,8 +12,18 @@ var _ = require('lodash'),
...
+        onlyCLIReporterEnabled,
+        skipExecutions;
+
+    // Need to do add this check for unit-tests when no options are passed
+    if (options) {
+        onlyCLIReporterEnabled = (_.isArray(options.reporters) &&
+            options.reporters.length === 1 &&
+            options.reporters.includes('cli'));
+        // skipExecutions in case the option is specified, or only cli is specified as reporter
+        skipExecutions = _.get(options, 'skipExecutions') || onlyCLIReporterEnabled;
```
coditva, line 25: We've already checked that `options` is not falsy, we can just use `options.skipExecutions` here.

coditva, line 25: In case `skipExecutions` is specified, we can avoid computing `onlyCLIReporterEnabled`.

CHANGELOG.yaml
```diff
@@ -1,3 +1,7 @@
+master:
+    fixed bugs:
+     - GH-2509 Fix to not store memory heavy objects in summary instead give options
```
coditva, line 3: I feel this is more of an improvement, than a bug fix. We're adding a new flag! We should add it under `new features:`.
What do you think?

CHANGELOG.yaml
coditva: Nope, still seems a little off. Can you read the last few changelogs in the file for a reference?

test/unit/run-summary.test.js
```diff
@@ -48,6 +48,33 @@ describe('run summary', function () {
...
+        var options = {
+                skipExecutions: true
+            },
+            summary = new Summary(new EventEmitter(), options),
+            executions = summary.run.executions;
+
+        expect(summary).to.have.property('run');
+        expect(_.keys(summary.run).sort())
+            .to.eql(['stats', 'timings', 'executions', 'transfers', 'failures', 'error'].sort());
+        expect(executions, 'should have 0 executions').to.have.lengthOf(0);
```
coditva, line 61: We can also assert that executions is an array.

test/unit/run-summary.test.js
```diff
@@ -48,6 +48,33 @@ describe('run summary', function () {
...
+    });
+
+    it('should have empty executions array when cli reporter specified', function () {
```
coditva, line 64:
```suggestion
    it('should have empty executions array when only cli reporter specified', function () {
```

test/unit/run-summary.test.js
```diff
@@ -48,6 +48,33 @@ describe('run summary', function () {
...
+
+        expect(summary).to.have.property('run');
+        expect(_.keys(summary.run).sort())
+            .to.eql(['stats', 'timings', 'executions', 'transfers', 'failures', 'error'].sort());
+        expect(executions, 'should have 0 executions').to.have.lengthOf(0);
+    });
+
     describe('runtime event statistics', function () {
         it('should track relevant events', function () {
             var emitter = new EventEmitter(),
```
coditva, line 80: Can we add an integration test as well?

test/unit/run-summary.test.js
```diff
@@ -48,6 +48,33 @@ describe('run summary', function () {
...
+                skipExecutions: false,
+                reporters: ['cli']
+            },
+            summary = new Summary(new EventEmitter(), options),
+            executions = summary.run.executions;
+
+        expect(summary).to.have.property('run');
+        expect(_.keys(summary.run).sort())
+            .to.eql(['stats', 'timings', 'executions', 'transfers', 'failures', 'error'].sort());
+        expect(executions, 'should have 0 executions').to.have.lengthOf(0);
```
coditva, line 75: We can also assert that executions is an array.

coditva (changes requested):
Requested a few changes.

CHANGELOG.yaml
```diff
@@ -1,3 +1,8 @@
+master:
+    new features:
+     - >-
+        GH-2509 Added a new option `skip-executions-summary` in newman to skips addition of request,     response to summary executions report generated for specified reporter.
+
```
coditva, line 5: Seems to be some formatting issues here. Lots of space before `response to summary...`

README.md
```diff
@@ -303,6 +303,7 @@ return of the `newman.run` function is a run instance, which emits run events th
 | options.sslExtraCaCerts   | The path to the file, that holds one or more trusted CA certificates in PEM format.<br /><br />_Optional_<br />Type: `string` |
 | options.requestAgents     | Specify the custom requesting agents to be used when performing HTTP and HTTPS requests respectively. Example: [Using Socks Proxy](#using-socks-proxy)<br /><br />_Optional_<br />Type: `object` |
 | options.cookieJar     | One can optionally pass a CookieJar file path as `string` to this property and that will be deserialized using [`tough-cookie`](https://github.com/salesforce/tough-cookie). This property also accepts a `tough-cookie` CookieJar instance.<br /><br />_Optional_<br />Type: `object\|string` |
+| options.skipExecutionsSummary     | Skips addition of request, response to summary executions report generated for specified reporter. <br /><br />_This option is recommended to be used if response size for API is large or the API response is not required to be included in the summary report._ |
```
coditva, line 306: Let's also add whether it's optional and its type. See the previous options for reference.

coditva, line 306: We missed out adding it to the `newman run <collection-file-source> [options]` section.

lib/run/summary.js
```diff
@@ -12,8 +12,20 @@ var _ = require('lodash'),
...
+    // Need to do add this check for unit-tests when no options are passed
+    if (options) {
+        if (!options.skipExecutionsSummary) {
+            onlyCLIReporterEnabled = (Array.isArray(options.reporters) &&
+                options.reporters.length === 1 &&
+                options.reporters.includes('cli'));
+        }
+        // skipExecutionsSummary in case the option is specified, or only cli is specified as reporter
+        skipExecutionsSummary = options.skipExecutionsSummary || onlyCLIReporterEnabled;
+    }
```
coditva, line 28: Multiple levels of nested `if` blocks seems a little yukk 😬

One way to remove this can be to update the `options.skipExecutionsSummary` itself when it is not set by the user and it is implicit, like so:

```suggestion
    if (!options.skipExecutionSummary) {
    	// skipExecutionsSummary in case the option is specified, or only cli is specified as reporter
    	options.skipExecutionSummary = (Array.isArray(options.reporters) &&
        	options.reporters.length === 1 &&
            options.reporters.includes('cli'));
    }
```

This is just one way, I'm open to more creative/better solutions. 🙏

bin/newman.js
```diff
@@ -60,6 +60,8 @@ program
     .option('--ssl-extra-ca-certs <path>', 'Specify additionally trusted CA certificates (PEM)')
     .option('--cookie-jar <path>', 'Specify the path to a custom cookie jar (serialized tough-cookie JSON) ')
     .option('--export-cookie-jar <path>', 'Exports the cookie jar to a file after completing the run')
+    .option('-s, --skip-executions-summary',
```
coditva, line 63: As we had discussed in the sync-up, we might need to change the option name.

test/cli/skip-execution-summary-options.test.js
```diff
@@ -0,0 +1,58 @@
...
+                    expect(jsonReport).to.have.property('run');
+                    expect(_.keys(jsonReport.run).sort())
+                        .to.eql(['stats', 'timings', 'executions', 'transfers', 'failures', 'error'].sort());
+                    expect(executions).to.be.an('array');
+                    expect(executions, 'should have 1 executions').to.have.lengthOf(1);
+                    done();
+                });
+            });
+    });
+});
```
coditva, line 58: Awesome! 💯

coditva: Can we add a similar test in `test/library/` as well for the sake of completeness?

CHANGELOG.yaml
lucifercr07: Fixed

README.md
lucifercr07: Fixed

lucifercr07: Fixed

bin/newman.js
lucifercr07: Changed to `--no-executions-summary` as discussed with @codenirvana

test/cli/skip-execution-summary-options.test.js
lucifercr07: Added

CHANGELOG.yaml
```diff
@@ -1,7 +1,8 @@
 master:
     new features:
      - >-
-        GH-2509 Added a new option `skip-executions-summary` in newman to skips addition of request,     response to summary executions report generated for specified reporter.
+        GH-2509 Added a new option `skip-executions-summary` in newman to skips addition of request,
```
coditva, line 4:
```suggestion
        GH-2509 Added a new option `no-executions-summary` in newman to skip addition of request,
```

README.md
```diff
@@ -303,7 +306,7 @@ return of the `newman.run` function is a run instance, which emits run events th
...
 | options.requestAgents     | Specify the custom requesting agents to be used when performing HTTP and HTTPS requests respectively. Example: [Using Socks Proxy](#using-socks-proxy)<br /><br />_Optional_<br />Type: `object` |
 | options.cookieJar     | One can optionally pass a CookieJar file path as `string` to this property and that will be deserialized using [`tough-cookie`](https://github.com/salesforce/tough-cookie). This property also accepts a `tough-cookie` CookieJar instance.<br /><br />_Optional_<br />Type: `object\|string` |
-| options.skipExecutionsSummary     | Skips addition of request, response to summary executions report generated for specified reporter. <br /><br />_This option is recommended to be used if response size for API is large or the API response is not required to be included in the summary report._ |
+| options.executionsSummary     | Skips addition of request, response to summary executions report generated for specified reporter. <br /><br /><br />Optional<br />_This option is recomended to be used if response size for API is large or the API response is not required to be included in the summary report._ |
```
coditva, line 309: We still haven't added the **Type** for the option (`boolean`).

lib/run/summary.js
```diff
@@ -13,19 +13,20 @@ var _ = require('lodash'),
...
-                options.reporters.includes('cli'));
-        }
-        // skipExecutionsSummary in case the option is specified, or only cli is specified as reporter
-        skipExecutionsSummary = options.skipExecutionsSummary || onlyCLIReporterEnabled;
+        executionSummaryOptions = _.get(options, 'executionsSummary', true),
+        skipExecutionSummary;
+
+    if (!executionSummaryOptions) {
+        // set skipExecutionSummary to true in case `--no-execution-summary` option is specified
+        skipExecutionSummary = true;
```
coditva, line 21: I think we can still get rid of this `if` block.

test/library/no-execution-summary-options.test.js
```diff
@@ -0,0 +1,66 @@
...
+    beforeEach(function () {
+        sh.test('-d', outDir) && sh.rm('-rf', outDir);
+        sh.mkdir('-p', outDir);
+    });
+
+    afterEach(function () {
+        sh.rm('-rf', outDir);
+    });
+
+    it('should correctly generate the json report with skip-executions option enabled', function (done) {
```
coditva, line 19: I don't think we're testing JSON reporter here.

```suggestion
describe('executionSummary option', function () {
    var outDir = 'out',
        outFile = outDir + '/newman-report.json';

    beforeEach(function () {
        sh.test('-d', outDir) && sh.rm('-rf', outDir);
        sh.mkdir('-p', outDir);
    });

    afterEach(function () {
        sh.rm('-rf', outDir);
    });

    it('when set should add executions in the result', function (done) {
```

coditva (changes requested):

test/cli/no-execution-summary-options.test.js
```diff
@@ -0,0 +1,58 @@
+var fs = require('fs'),
+
+    _ = require('lodash'),
+    sh = require('shelljs');
+
+describe('JSON reporter', function () {
```
coditva, line 6: Similar to the comment to `test/library/` test, we should rename the `describe` and `it` blocks here.

test/library/no-execution-summary-options.test.js
```diff
@@ -0,0 +1,66 @@
+var fs = require('fs'),
```
coditva, line 1: These tests should be added inside `test/library/run-options.test.js`, I feel.

lib/run/summary.js
lucifercr07: Combining the `cli` only present as reporter check with the `!executionSummaryOptions` we can eliminate the block but I guess it'll make the code less readable. Please let me know if you've any other suggestions for this.

README.md
lucifercr07: Fixed

CHANGELOG.yaml
lucifercr07: Fixed

test/cli/no-execution-summary-options.test.js
lucifercr07: Fixed

lib/run/summary.js
coditva: How about this: https://github.com/postmanlabs/newman/pull/2509#discussion_r493198231

Single `if` block without combining.

coditva (approved):

coditva:
LGTM! 🚀
Few tests failing, verify that they pass and I think we're good to go then.

test/cli/no-execution-summary-options.test.js
```diff
@@ -0,0 +1,58 @@
...
+    beforeEach(function () {
+        sh.test('-d', outDir) && sh.rm('-rf', outDir);
+        sh.mkdir('-p', outDir);
+    });
+
+    afterEach(function () {
+        sh.rm('-rf', outDir);
+    });
+
+    it('when executionsSummary option specified should not add executions in the result', function (done) {
```
codenirvana, line 19: Improve the test description.

test/library/run-options.test.js
```diff
@@ -239,4 +241,66 @@ describe('Newman run options', function () {
             });
         });
     });
+
```
codenirvana, line 244: - add test for default behavior, without `reporters` and `executionsSummary` option
- add a test with `reporters: 'cli'`

test/unit/run-summary.test.js
```diff
@@ -1,4 +1,5 @@
 var _ = require('lodash');
+const { expect } = require('chai');
```
codenirvana, line 2: Why it's not needed before? and why it's mixed with `var` and `const`?

test/unit/run-summary.test.js
```diff
@@ -48,6 +49,35 @@ describe('run summary', function () {
         expect(summary.run.transfers).to.be.an('object');
     });
 
+    it('should have empty executions array', function () {
```
codenirvana, line 52: should have empty executions array when......?

test/unit/run-summary.test.js
```diff
@@ -48,6 +49,35 @@ describe('run summary', function () {
...
+        var options = {
+                executionsSummary: false
```
codenirvana, line 54: Add test for `executionsSummary`:
- false
- true
- undefined (default behavior, don't set `undefined` explicitly)

codenirvana: And another set of tests for how it changes based on the `reporters` option.

codenirvana (changes requested):

lib/run/summary.js
```diff
@@ -12,7 +12,15 @@ var _ = require('lodash'),
...
     // keep a copy of this instance since, we need to refer to this from various events
-    var summary = this;
+    var summary = this,
+        skipExecutionsSummary = !_.get(options, 'executionsSummary', true);
+
+    if (!skipExecutionsSummary) {
+        // skipExecutionsSummary in case the option is specified, or only cli is specified as reporter
+        skipExecutionsSummary = (options && Array.isArray(options.reporters) &&
+            options.reporters.length === 1 &&
+            options.reporters.includes('cli'));
```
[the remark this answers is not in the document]
lucifercr07: Other than the callback method passed to newman run I can't find any other way to differentiate between the newman cli or library calls.
Is there any other way you can suggest to find the difference? I don't want to rely on some extra param passing for lib calls as it might break backward compatibility

coditva:
Pending test changes in `run-summary.test.js`.

@codenirvana does [this](https://github.com/postmanlabs/newman/pull/2509/files#diff-95cf23cc6a55e698a678e439f3c802a48774747e4560af3ed39314959ebf09abR80) seem like a valid approach for [skipping execution summary for CLI reporter in CLI mode](https://github.com/postmanlabs/newman/pull/2509/files#diff-029feab71a45a0c6f6d5fda9e94d34ac178a0ef29f61fc7594fc86a62f64002fR24)?

PR #2518, opened 2020-09-24
Publish types from DefinitelyTyped with newman package

jbaldassari:
I recently ran into a couple of problems when updating `newman` and `postman-collection` to their latest versions along with the latest versions of [@types/newman](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/newman) and [@types/postman-collection](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/postman-collection).

The first problem was that [@types/postman-collection](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/postman-collection) is no longer compatible with `postman-collection`.  For example, `VariableScopeDefinition` is now `VariableScope.definition`.  It appears that the type information is no longer being maintained in the DefinitelyTyped repo and has instead been moved inside the `postman-collection` repo [here](https://github.com/postmanlabs/postman-collection/blob/develop/types/index.d.ts).

The second problem was that `newman` does not publish types in its repo, so the only way to get them currently is from DefinitelyTyped.  However, the `newman` types in the DefinitelyTyped repo [depend](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/newman/index.d.ts#L10-L15) on the outdated `postman-collection` types in DefinitelyTyped, so they can't be used.

The only way to use the most recent versions of `newman` and `postman-collection` right now is to remove both `@types/newman` and `@types/postman-collection` from your project.  After doing this the `postman-collection` type information will be available directly from that project, but it is not possible to get `newman` types such as `NewmanRunSummary`.

In this PR I've taken the types from [@types/newman](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/newman) and added them to the `newman` repo in the same way as the `postman-collection` repo.  I did also try to replicate how `postman-collection` automatically generates these types with `npm/build-types.js`, but it looks like the many of the JSDocs need to be updated for compatibility with tsd-jsdoc.  So for now `lib/types/index.d.ts` comes directly from [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/newman/index.d.ts) with some minor changes for compatibility with `postman-collection` 3.6.7.  I've confirmed that with these changes I can access the `newman` types in my code.

I'm not sure what the strategy is for the postman/newman team regarding TypeScript, but if the goal is to publish types along with their respective packages (which I think is a good idea) then hopefully this PR might help move in the right direction.  Alternatively, if type support is already in progress in `newman` I could submit both the `newman` and `postman-collection` type changes to the DefinitelyTyped repo in the meantime.

### DIFF
--- a/lib/types/index.d.ts
+++ b/lib/types/index.d.ts
@@ -1,0 +1,206 @@
+// Type definitions for newman 5.1
+// Project: https://github.com/postmanlabs/newman
+// Definitions by: Leonid Logvinov <https://github.com/LogvinovLeon>
+//                 Graham McGregor <https://github.com/Graham42>
+// Definitions: https://github.com/DefinitelyTyped/DefinitelyTyped
+// TypeScript Version: 2.4
+
+import * as http from 'http';
+import { EventEmitter } from "events";
+import {
+    Collection,
+    VariableScope,
+} from "postman-collection";
+
+export interface NewmanRunOptions {
+    /** A JSON / Collection / String representing the collection. */
+    collection: Collection | Collection.definition | string;
+    /** An environment JSON / file path for the current collection run. */
+    environment?: VariableScope | VariableScope.definition | string;
+    /** A globals JSON / file path for the current collection run. */
+    globals?: VariableScope | VariableScope.definition | string;
+    /** The relative path to export the globals file from the current run to  */
+    exportGlobals?: string;
+    /** The relative path to export the environment file from the current run to */
+    exportEnvironment?: string;
+    /** The relative path to export the collection from the current run to */
+    exportCollection?: string;
+    /**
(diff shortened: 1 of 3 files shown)


codenirvana (commented):
@jbaldassari `postman-collection` v3.6.9 no longer enforce the auto-generated bundled type declarations which are causing all these issues.
So `@types/newman` and `@types/postman-collection` will work just fine. Refer https://github.com/postmanlabs/postman-collection/issues/602#issuecomment-753507973 for more details.

Moving forward we would like to keep types closer to both the packages but currently, we don't have the bandwidth to maintain the type declarations so closing this PR for now.

Thanks for your contribution and sorry for the delayed response.

Issue #2197, opened 2019-12-18
Junit report only contains results from last iteration


rycornell:
When using newman with a data source file, the Junit report only contains the results for the last item in the file.  The CLI output correctly shows the number of iterations, test scripts and assertions, while the Junit xml file only shows a single test case.

 1. Newman Version (can be found via `newman -v`): 4.5.7
 2. OS details (type, version, and architecture): Windows 10 64bit
 3. Are you using Newman as a library, or via the CLI? CLI
 3. Did you encounter this recently, or has this bug always been there: Always
 4. Expected behaviour: The JUnit report should contain results for all iterations.
 5. Command / script used to run Newman: newman run collection.json -d data.csv -r cli,junit
 6. Sample collection, and auxiliary files (minus the sensitive details): The sample collection has 1 request: https://postman-echo.com/get?iter={iter}
 7. Screenshots (if applicable):

![image](https://user-images.githubusercontent.com/3660285/71096715-cb60c700-217c-11ea-9366-fdcd75b47db0.png)
![image](https://user-images.githubusercontent.com/3660285/71096784-e8959580-217c-11ea-91fa-e11d6484d45f.png)
![image](https://user-images.githubusercontent.com/3660285/71097033-5f329300-217d-11ea-8170-bb2d3a22c865.png)
![image](https://user-images.githubusercontent.com/3660285/71097106-796c7100-217d-11ea-86c6-91c240581451.png)

jander99:
Also seeing this issue with 101 iterations only yielding one testsuite.

Christian0411:
Hey I was able to get all the iterations to report by using newman-reporter-junitfull. You can get it via npm 
npm install -g newman-reporter-junitfull

rycornell:
This is perfect, thank you! @Christian0411

twoodhouse:
@rycornell, can we reopen this? Adding a plugin as a dependency doesn't fix the issue internal to newman. The standard junit reporter should have been smart enough to handle this.

adityaofficial10:
Is anyone already working on this issue?
If not, I am ready to start.

adityaofficial10:
#2601

adityaofficial10:
This is the same as Issue #2601 
That was for json reporter.
@DannyDainton 
Merging #2614 will make it easy for me to open a PR for this issue.

sxmartin:
[...]

PR #2614, opened 2021-02-07
Fix:Timing Details added for each request(#2601)

### LINKED ISSUE #2601: output request timing details for each request
not sure if there is a way to output request timing details (dns, firstBytes .....) for each request. I can only find timing details for the whole collection. For each request, only response time is available in the output file.

 1. Newman Version (can be found via `newman -v`): 5.2.0
 3. Are you using Newman as a library, or via the CLI? cli
 4. Expected behaviour:

adityaofficial10:
Output timing details( like dns, firstBytes,..) for each request are now, there in the report using JSON reporter and verbose option.
Fixes #2601 

The fix details are as follows:
1. A separate attribute requestWise has been added to timings attribute of summary.run .
2. Timing details like dns, firstByte, wait, prepare, etc have been added for each request, whenever verbose option is set.

Background of the issue:
Actually, CLI reporter gives all the details using the verbose option but JSON reporter only publishes the overall timing details in the report.

### DIFF
--- a/lib/run/summary.js
+++ b/lib/run/summary.js
@@ -267,7 +267,11 @@ _.assign(RunSummary, {
     attachTimingTrackers (summary, emitter) {
         // mark the point when the run started
         // also mark the point when run completed and also store error if needed
-        emitter.on('start', function () { summary.run.timings.started = Date.now(); });
+        // initialise a requestWise timings array
+        emitter.on('start', function () {
+            summary.run.timings.started = Date.now();
+            summary.run.timings.requestWise = [];
+        });
         emitter.on('beforeDone', function () {
             summary.run.timings.completed = Date.now();
         });
@@ -319,6 +323,11 @@ _.assign(RunSummary, {
             timings = timings && timings.timings;
             timingPhases = timings && sdk.Response.timingPhases(timings);
 
+            // Push the timing stats of the current request to the summary
+            summary.run.timings.requestWise && summary.run.timings.requestWise.push(timingPhases);
+
+            // Computing the timing stats for the whole run
+            // eslint-disable-next-line lodash/collection-method-value
             (timingPhases || time) && _.forEach([
                 'dns',
                 'firstByte',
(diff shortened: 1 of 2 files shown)


adityaofficial10:
@codenirvana I think I can refine this idea even more by adding iteration count and item name to the mentioned array.
That would also help us to close another issue #2197

adityaofficial10:
@DannyDainton @codenirvana I've added the iteration data and item name as well in a separate commit.
This will help us in closing #2197

DannyDainton:
Hey @adityaofficial10

The best person to tag is @vvs11 🏆

codenirvana (changes requested):
A better approach would be to add `timingPhases` to each `run.executions`.

adityaofficial10:
@codenirvana 
Sure, I agree with you..
I'll follow up with the changes shortly.
Thanks for the review.

adityaofficial10:
@codenirvana 
I found that attachTimingTrackers is called before attachReportingTrackers.
So implementing this would make the codeflow complex.
Could you please guide me regarding this?

lib/run/summary.js
```diff
@@ -319,14 +324,19 @@ _.assign(RunSummary, {
...
+            // Computing the timing stats for the whole run
             (timingPhases || time) && _.forEach([
                 'dns',
                 'firstByte',
                 'response'
             ], (value) => {
                 var currentValue = (value === 'response') ? time : (timingPhases && timingPhases[value]),
                     previousAverage = summary.run.timings[`${value}Average`],
-                    previousVariance = Math.pow(summary.run.timings[`${value}Sd`], 2),
+                    previousVariance = summary.run.timings[`${value}Sd`],
```
shamasis, line 339: Sorry for my ignorance here. Can you expand what this change was for?

shamasis (commented):
Why not use timings from the executions array itself?

shamasis (changes requested):
Why not use timings from the executions array itself and attach the same in JSON reporter?

I am not in favour of saying that the "summary" now has individual data points instead of containing summary.

TheRealDashu:
So is there any progress here ? I dont see anything that compares to the beautifull requestWise field that was composed here in the timings. In the execution array. There is no wait time no prepare time nothing that compares to the details shown in this proposal.

Issue #2465, opened 2020-08-10
Command line arg to exclude folders


dcmeglio:
1. Newman Version (can be found via `newman -v`): 5.1.0
 2. OS details (type, version, and architecture): Mac OS X 19.3.0
 3. Are you using Newman as a library, or via the CLI? CLI

Really more of a suggestion than an issue. It would be nice if newman could have a --exclude-folder that is the OPPOSITE of --folder

postman.setNextRequest can be used to simulate this, however it requires that tests be in a very particular order. If someone reorders tests they might not realize that that there are such statements in a pre-request script and now we skip the wrong tests.

It would be nice if from newman we could exclude particular folders so that if there are conditional tests, we can simply put them in those folders and alter command line args as needed to include/exclude.

adityaofficial10:
@DannyDainton Are we looking forward to this request?
If yes, I can work on this as this is similar to another feature request related to options.

DannyDainton:
Hey @adityaofficial10

The best person to tag is @vvs11 🏆

adityaofficial10:
> Hey @adityaofficial10
> 
> The best person to tag is @vvs11 🏆

Sure, thanks for guidance.

daniavander:
when can we start to use it?

PR #2620, opened 2021-02-13
Feature:Option --exclude-folder added to the options.

adityaofficial10:
**What this PR does?**
This PR adds an option --exclude-folder to the list of options.
This is in response to the feature request #2465 

- There are changes to the main newman file and the run/options.js file in particular, with --exclude-folder option added.
- Tests have been added to test the option, i.e CLI tests, Unit tests, Integration tests and Library Tests.

**What is the added option --exclude-folder all about?**
- The option --exclude-folder can take in one or more than one folders and it excludes them from items to be executed.
- This option rules supreme in case same folder has been passed using both --folder and --exclude-folder options.

### DIFF
--- a/bin/newman.js
+++ b/bin/newman.js
@@ -28,6 +28,9 @@ program
     .option('--folder <path>',
         'Specify the folder to run from a collection. Can be specified multiple times to run multiple folders',
         util.cast.memoize, [])
+    .option('--exclude-folder <path>',
+        'Specify the folder to exclude from a collection. Can be specified multiple times to exclude multiple folders',
+        util.cast.memoize, [])
     .option('--global-var <value>',
         'Allows the specification of global variables via the command line, in a key=value format',
         util.cast.memoizeKeyVal, [])
--- a/lib/run/index.js
+++ b/lib/run/index.js
@@ -96,6 +96,10 @@ module.exports = function (options, callback) {
     var emitter = new EventEmitter(), // @todo: create a new inherited constructor
         runner = new runtime.Runner(),
         stopOnFailure,
+        items,
+        itemsLength,
+        excludedItems = [],
+        i,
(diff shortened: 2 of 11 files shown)


adityaofficial10:
@codenirvana @vvs11
This was a feature request actually.
So I'd like to hear from you guys on this.

codenirvana (commented):
@adityaofficial10 This is great but I am not sure about mutating the collection here. 🤔 

Also, why just folders? What about excluding requests?

adityaofficial10:
> @adityaofficial10 This is great but I am not sure about mutating the collection here. 🤔
> 

Actually @codenirvana ,
We use entrypoint for folders option and imitating something similar would require a change to the postman-runtime.
So mutating was a better option according to me..
It would be really great if you could guide me regarding the correct way.
> Also, why just folders? What about excluding requests?
Also, I am working on that in my next patch.
I'll follow up with that shortly..

adityaofficial10:
@codenirvana @shamasis @coditva 
For requests, I think of passing a path to the request like folder1/folder2/request1 and then using search mechanism to exclude it.
Does this look good?

rozanecm:
@adityaofficial10 I think the best way to implement request/folder skipping would be maintaining symmetry with the `--folder` option, in which one can specify a folder name or a request name directly.

Having said that, I do think it'd be nice to be able to specify full paths to folders/requests rather than just the name. This has been brought up several times in the forums throughout the years (for the `--folder` case) and has never been addressed.

pixincreate:
Hey there, it's been 2+ years and the PR is still open, may I know any ETA for this?  
Is there any alternative for this until then? I've currently written a function that fetches all the folder names excluding the name the folder that needs to be ignored and this has a lot of overhead.

PR #2627, opened 2021-02-22
Update license year to 2021

PawelBorkar:
Update the license year to 2021 from 2016


codenirvana (commented):
Thanks for the contribution.
Although these licenses can have multiple years or the current year but, we would like to keep it as first published year i.e, 2016.

PR #2630, opened 2021-02-23
Feature: Access Iteration wise Request Data from Summary

### LINKED ISSUE #2103: unable to access iterationData from summary object
1. Newman Version (can be found via `newman -v`): 4.5.4
 2. OS details (type, version, and architecture): Windows 10 
 3. Are you using Newman as a library, or via the CLI? Issue affects both
 3. Did you encounter this recently, or has this bug always been there: Not sure (started using newman recently)
 4. Expected behaviour: NA - feature request
 5. Command / script used to run Newman: NA
 6. Sample collection, and auxiliary files (minus the sensitive details): BA
 7. Screenshots (if applicable): NA

Currently I can see no way of accessing the iteration data (data from the csv file) after the completion of the execution. 

It will be useful to get access to this so that we can use this data in the reports . It is reasonable to expect to report/log the data values used for a specific test run/iteration.

It will be nice to be able to access this via the summary object like this : summary.iterationData[iterationNumber].variableName

Is there any way to get to this data from the summary object as of now?

adityaofficial10:
Fixes #2103 
**What does this PR do?**
This PR is based on a feature request #2103 and it adds the functionality to access iteration data from the summary itself.

- An array iterationData has been added to the summary.run object.
- The array iterationData consists of objects which are of the type { iteration: iterationNumber, requestTimings: [Objects] }.
- The requestTimings array is an array of objects consisting of the timing details of each request in a particular iteration.

**Background of the Request:**
The request was about adding the functionality that would allow access to iteration wise data from the summary itself.


adityaofficial10:
@codenirvana 
I would like the community to review this.

shamasis (changes requested):
This approach causes heavy memory pressure on Newman runs.

adityaofficial10:
@shamasis 
I would request you to open this.
We can try another approach.

shamasis:
Check if reporters can already access newman commander options. If yes, then nothing more needs to be built. Reporters then already have access to iteration data file name.

adityaofficial10:
@shamasis 
You're right.
This is not needed.
Thanks again for your valuable feedback.

PR #2639, opened 2021-03-02
Not rely on CI=true for TTY check

### LINKED ISSUE #2638: Newman --color auto should give precedence to TERM over CI=true
Newman should give precedence to `TERM` environment variable, even if `CI=true` ([Jenkins supports colors](https://plugins.jenkins.io/ansicolor/)).

With `TERM` and `CI=true`:
image

With `TERM` and without `CI=true`:
image

felipecrs:
Fixes #2638

### DIFF
--- a/lib/reporters/cli/cli-utils.js
+++ b/lib/reporters/cli/cli-utils.js
@@ -93,7 +93,7 @@ cliUtils = {
      * color:     |  noTTY:
      * 'on'      -> false
      * 'off'     -> true
-     * otherwise -> Based on isatty.
+     * otherwise -> Based on isTTY.
      *
      * @param {String} color - A flag to indicate usage of the --color option.
      * @returns {Boolean} - A boolean value depicting the result of the noTTY check.
--- a/lib/reporters/cli/cli-utils.js
+++ b/lib/reporters/cli/cli-utils.js
@@ -93,14 +93,13 @@ cliUtils = {
      * color:     |  noTTY:
      * 'on'      -> false
      * 'off'     -> true
-     * otherwise -> Based on CI or isTTY.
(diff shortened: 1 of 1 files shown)


lib/reporters/cli/cli-utils.js
```diff
@@ -93,14 +93,13 @@ cliUtils = {
...
-     * otherwise -> Based on CI or isTTY.
+     * otherwise -> Based on isatty.
      *
      * @param {String} color - A flag to indicate usage of the --color option.
      * @returns {Boolean} - A boolean value depicting the result of the noTTY check.
      */
     noTTY: function (color) {
-        return (color === 'off') || (color !== 'on') &&
-            (Boolean(process.env.CI) || !process.stdout.isTTY); // eslint-disable-line no-process-env
+        return (color === 'off') || (color !== 'on') && (!require('tty').isatty(1));
```
codenirvana, line 102: What's the difference between `process.stdout.isTTY` and `require('tty').isatty(1)`? 🤔 

I feel this will have the same behaviour:
```suggestion
        return (color === 'off') || (color !== 'on') && (!process.stdout.isTTY);
```

codenirvana (changes requested):

lib/reporters/cli/cli-utils.js
felipecrs: I searched some time ago and I found that isTTY not always works as good as isatty works across all the different node versions. I'll try to find it again, and let you know.

felipecrs: Here it is:

https://github.com/babel/babel/issues/3763
https://github.com/nodejs/node/issues/28386

codenirvana (approved):

Issue #2560, opened 2020-11-17
[Question] Newman - Dynamic Iteration with Data File


Xdomen:
Hey guys,

I have a question regarding newman usage with CSV Data files.

We have about 150 tests within a single CSV file, however, we would like to call them dynamically via parameter. Meaning if I want to initiate test iteration 149, I would have to make 148 iteartions prior to that test. This is causing us to have multiple CSV files whic his quite unnecessary. 

**Example of CSV Data file:**

```
url, user_id, type, amount,
{{baseUrl}}, 1, 0, 100
{{baseUrl}}, 2, 1, 200
{{baseUrl}}, 2, 1, 300
```

Each row in CSV is an individual test and counts as an iteration. How could we initiate tests dyniamically i.e execute third test iteration without having execute second and third test iteration? 

Any help on this is highly appreciated!

Thanks.

praveendvd:
Really a good feature , i also feel this hectic i have to create new temp csv to test  a sngle line

adityaofficial10:
@Xdomen 
I'm moving forward with this feature request.

adityaofficial10:
@preyunk
I was working with this and had my draft ready.
You should have informed me about your work atleast.
Anyways, I have the tests ready.
So we can collaborate I guess.

preyunk:
Yeah sure!
Sorry for the inconvenience, though.

Xdomen:
@adityaofficial10 could you please help us with the opened PR https://github.com/postmanlabs/newman/pull/2652 🙏

PR #2652, opened 2021-03-11
Add --iteration-data-range option

preyunk:
This option enables dynamic iteration of data file
this option will accept start index and an optional end index to
select a range from start to end - 1.

Fixes: https://github.com/postmanlabs/newman/issues/2560


preyunk:
This will probably need some tests for invalid values of start and end index

adityaofficial10:
@preyunk 
I guess you'll need some tests to ensure proper coverage.

preyunk:
@shamasis @codenirvana 
Can you take a look if the approach to implementing the feature is correct?

codenirvana (changes requested):
I feel there's a need for `from` and `to` instead of asking for the complete `range`.

Also, the count should be 1-based (instead of 0).

Xdomen:
@preyunk can I somehow help with the outstanding comments? :)

preyunk:
> @preyunk can I somehow help with the outstanding comments? :)

Yeah, sure 😀

Xdomen:
@preyunk seems like I bit more than I could chew, I don't know how to make the necessary changes, so If you could take over I would be really grateful 🙏  Sincere aplogies for the inconvenience..

shamasis (changes requested):

All the above seems off. These should be done at runtime level. Instead of manipulating data file, runtime itself should start and end on specific iterations. It would also help it make portable across CLI and App

PR #2654, opened 2021-03-11
Feature: Running Single Request from a folder in a Collection

### LINKED ISSUE #2566: [Feature Request] Enable running single request from the folder/collection
1. Newman Version (can be found via newman -v): 5.2.1
2. OS details (type, version, and architecture): Linux
3. Are you using Newman as a library, or via the CLI? CLI
4. Did you encounter this recently, or has this bug always been there: Always

1. newman run collection.json -e dev_env.json --folder RequestFolder/PostRequest

Expected behaviour:
- should run the single request

adityaofficial10:
Fixes #2566.

**What does this PR do?**
This PR is a follow-up to the feature request mentioned under the above cited issue.
It allows the user to specify a path to an item which doesn't have unique name in the collection and runs the item at the specified path.

- An extra option --item-from-folder has been added.
- It can have one or more than one arguments.
- It uses string for one argument and array for more than one arguments.
- It overrides folder option if both are specified.
- It halts the run if the path specified is invalid.
- Unit, Library and Integration Tests have been added.

**What things are still expected?**
Since I am still working on it, there will be following changes:
- Currently, only paths of depth 2 are supported, i.e myFolder/myRequest. Functionality for deeper paths is being implemented.
- Functionality for more than one arguments is being implemented.
- Some more unit tests will be added.
- CLI tests will be added.
P.S. Coverage may not be appropriate right now but will increase surely once remaining CLI and Unit tests are added.

### DIFF
--- a/bin/newman.js
+++ b/bin/newman.js
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 
+
 require('../lib/node-version-check'); // @note that this should not respect CLI --silent
 
 const _ = require('lodash'),
@@ -95,6 +96,15 @@ program.on('command:*', (command) => {
     console.error(`error: invalid command \`${command}\`\n`);
     program.help();
 });
+
+program
+    .command('run-request <url>')
+    .usage('<url> [options]')
+    .option('-m,--method', 'Method of the request', 'GET')
+    .action((url, command) => {
+        console.info(url);
+        console.info(command);
+    });
 
 /**
  * Starts the script execution.
--- a/bin/newman.js
+++ b/bin/newman.js
@@ -29,6 +29,8 @@ program
     .option('--folder <path>',
         'Specify the folder to run from a collection. Can be specified multiple times to run multiple folders',
         util.cast.memoize, [])
+    .option('--item-from-folder <path>', 'Specify the item and its parent folder',
+        util.cast.memoize, [])
     .option('--global-var <value>',
         'Allows the specification of global variables via the command line, in a key=value format',
         util.cast.memoizeKeyVal, [])
--- a/lib/run/index.js
+++ b/lib/run/index.js
@@ -136,13 +136,19 @@ module.exports = function (options, callback) {
         }
 
         // sets entrypoint to execute if options.folder is specified.
-        if (options.folder) {
+        if (options.folder && !options.itemFromFolder) {
             entrypoint = { execute: options.folder };
 
             // uses `multipleIdOrName` lookupStrategy in case of multiple folders.
             _.isArray(entrypoint.execute) && (entrypoint.lookupStrategy = MULTIENTRY_LOOKUP_STRATEGY);
         }
 
+        if (options.itemFromFolder) {
+            if (options.collection.items) {
+                options.collection.items.members = util.parseItemFromFolder(String(options.itemFromFolder),
+                    options.collection.items.members);
+            }
+        }
         // sets stopOnFailure to true in case bail is used without any modifiers or with failure
         // --bail => stopOnFailure = true
         // --bail failure => stopOnFailure = true
--- a/lib/run/options.js
+++ b/lib/run/options.js
@@ -256,6 +256,26 @@ var _ = require('lodash'),
         },
 
         /**
+         * Helper function to sanitize folder option.
+         *
+         * @param {String[]|String} value - The list of folders to execute
+         * @param {Object} options - The set of wrapped options.
+         * @param {Function} callback - The callback function invoked to mark the end of the folder load routine.
+         * @returns {*}
+         */
+        itemFromFolder: function (value, options, callback) {
+            if (!value.length) {
+                return callback(); // avoids empty string or array
+            }
+
+            if (Array.isArray(value) && value.length === 1) {
+                return callback(null, value[0]); // avoids using multipleIdOrName strategy for a single item array
+            }
+
+            callback(null, value);
+        },
+
+        /**
          * The iterationData loader module, with support for JSON or CSV data files.
          *
          * @param {String|Object[]} location - The path to the iteration data file for the current collection run, or
--- a/lib/util.js
+++ b/lib/util.js
@@ -252,6 +252,69 @@ util = {
             });
     },
 
+
+    /**
+     * Finds an item or item group based on id or name.
+     *
+     * @param {Object} itemGroup - The group of items to be searched within.
+     * @param {?String} match - The string name of the item to be compared against.
+     *
+     * @returns {Object} - The item we searched for.
+     */
+    findItemOrGroup: function (itemGroup, match) {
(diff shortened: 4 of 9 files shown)


adityaofficial10:
@codenirvana 
Once I add the tests here, this feature can be really useful as it would make Newman more versatile with respect to choice of items. It can actually be considered as an upgrade to folder option.
I might merge the new option with folder option if it is possible.

bin/newman.js
```diff
@@ -28,6 +29,8 @@ program
     .option('--folder <path>',
         'Specify the folder to run from a collection. Can be specified multiple times to run multiple folders',
         util.cast.memoize, [])
+    .option('--item-from-folder <path>', 'Specify the item and its parent folder',
```
shamasis, line 32: Use `request` vocabulary instead of item please. 😊

Also the parameter name is not sitting well. Postman runner has a way to sub-select which requests to run and this should feel close to that.

bin/newman.js
```diff
@@ -96,6 +99,15 @@ program.on('command:*', (command) => {
     program.help();
 });
 
+program
+    .command('run-request <url>')
```
shamasis, line 103: This change does not seem atomic or aligned with the original theme of the PR

This also conflicts with the upcoming `Newman request` command.

I'm guessing we should remove this.

lib/run/index.js
```diff
@@ -136,13 +136,19 @@ module.exports = function (options, callback) {
         }
 
         // sets entrypoint to execute if options.folder is specified.
-        if (options.folder) {
+        if (options.folder && !options.itemFromFolder) {
```
shamasis, line 139: This makes the two options mutually exclusive. Let users provide a folder and requests too. 

If both provided then the supplied request command will add the folder and the selected requests.

lib/run/index.js
```diff
@@ -136,13 +136,19 @@ module.exports = function (options, callback) {
...
             entrypoint = { execute: options.folder };
 
             // uses `multipleIdOrName` lookupStrategy in case of multiple folders.
             _.isArray(entrypoint.execute) && (entrypoint.lookupStrategy = MULTIENTRY_LOOKUP_STRATEGY);
         }
 
+        if (options.itemFromFolder) {
```
shamasis, line 146: Revise code. Block nesting not needed here.

lib/run/index.js
```diff
@@ -136,13 +136,19 @@ module.exports = function (options, callback) {
...
+            if (options.collection.items) {
```
shamasis, line 147: This is a bit risky. Direct manipulation of collection SDK's internal members object leads to inconsistencies.

lib/run/options.js
```diff
@@ -255,6 +255,26 @@ var _ = require('lodash'),
...
 
+        /**
+         * Helper function to sanitize folder option.
+         *
+         * @param {String[]|String} value - The list of folders to execute
+         * @param {Object} options - The set of wrapped options.
+         * @param {Function} callback - The callback function invoked to mark the end of the folder load routine.
+         * @returns {*}
+         */
+        itemFromFolder: function (value, options, callback) {
```
shamasis, line 266: There should be collection SDK built in functions for this. Double check once.

lib/util.js
```diff
@@ -252,6 +252,69 @@ util = {
...
+
+    /**
+     * Finds an item or item group based on id or name.
+     *
+     * @param {Object} itemGroup - The group of items to be searched within.
+     * @param {?String} match - The string name of the item to be compared against.
+     *
+     * @returns {Object} - The item we searched for.
+     */
+    findItemOrGroup: function (itemGroup, match) {
```
shamasis, line 264: There should be collection SDK built in find functions for this. Check once.

shamasis (changes requested):
This PR seems to need some changes. Inline comments added. 

Also the overall approach might need change since the filtering mechanism seems to be very invasive. 

Try iterating over the collection and removing unwanted items and item groups and then letting the collection flow through. @codenirvana would want your second opinion before proceeding.

adityaofficial10:
@shamasis 
Actually I thought over that and thought that it would increase running time but now I actually think this is the wrong way...
Thanks for your valuable feedback.
I'll make the required changes...

shamasis:
Double check @codenirvana once. I know there are some concerns about mutating original collection. A cloning may be needed. Or there is an underlying runtime extraction option that can be used for this.

adityaofficial10:
@shamasis 
Actually @codenirvana told me to find a different way if I can, in a similar pull request.
He's not sure about mutating the collection object..

shamasis:
Let's wait on @codenirvana for feedback. 

See, mutating collection is a problem only because in programmatic usage and export scenarios the user is not expecting the collection object to unexpectedly change. However, if we clone and mutate it solves the input worries (though adds memory pressure) and for output, it's a matter of choice and documentation (and reviewing the side-effects.)

adityaofficial10:
@shamasis 
I understand. 
Thanks for the review.

bin/newman.js
adityaofficial10: @shamasis 
It crept in by mistake.
I was working on the request command as well.😅

adityaofficial10:
@shamasis 
I've found a way out here.
We can actually take some code from extractRunnableItems and use it here. It is basically a recursive method to reach any folder depth, if the folder exists.
I'll move ahead with this.

adityaofficial10:
@codenirvana @shamasis 
Kindly guide me regarding how can I make both folder and this path to a request option work together.

codenirvana:
@adityaofficial10 You just have to find the specified request's `item.id` and use the `lookupStrategy=idOrName` but here the challenge is, what's the best way to take this (path leading to a request) input?

Honestly, I don't feel we need to introduce this complex CLI input at this moment because one can execute a single request using `--folder` also, provided the request name is unique in the request (otherwise it choose the first matching request).

And to overcome this limitation, you can pass the request UUID... which is very easy to get with Postman v8's info panel:

Screenshot 2021-04-04 at 5 08 27 PM

adityaofficial10:
@codenirvana 
Thanks for the help.
What I feel is passing a path like ' parent/child ' would work and it's not very complex too as it can be interpreted by the end users...

adityaofficial10:
@codenirvana 
I've found out a way.
For qualified path requests, we can just fetch the UUID and pass it to the entrypoint just like we do with folders.
This way both of them will work together.
I feel this feature can be really great..
Kindly allow me to go ahead..

bin/newman.js
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env node
 
+
```
gfernandes410, line 3: I think this line is not necessary.

gfernandes410:
@adityaofficial10 are you still working on this PR?

PR #2657, opened 2021-03-12
Newman Command for Single Request

raghavbhatia336:
Proof of concept to create `newman request` according to following expected output:

1. Users can send a single request with Newman and show the request/response using a reporter of their choice.
2. Replacing curl with newman request in a curl command should run the send the request using Newman.

Supported commands (The following commands show expected output in POC)

```
newman request -X GET http://www.google.com -r cli,json

newman request -X POST  https://postman-echo.com/post -H 'Content-Type: text/plain'  -r progress

```

### DIFF
--- a/bin/newman.js
+++ b/bin/newman.js
@@ -85,6 +85,74 @@ program
         });
     });
 
+// The `request` command allows you to run single request with the provided options.
+program
+    .command('request <url>')
+    .description('Run single request with provided options')
+    .usage('<url> [options]')
+    .option('--request, -X <method>', 'Specify request method', 'GET')
+    .option('--header, -H <header>', 'Specify request header', util.cast.memoize,  [])
+    .option('-e, --environment <path>', 'Specify a URL or path to a Postman Environment')
+    .option('-g, --globals <path>', 'Specify a URL or path to a file containing Postman Globals')
+    .option('-r, --reporters [reporters]', 'Specify the reporters to use for this run', util.cast.csvParse, ['cli'])
+    .option('-n, --iteration-count <n>', 'Define the number of iterations to run', util.cast.integer)
+    .option('-d, --iteration-data <path>', 'Specify a data file to use for iterations (either JSON or CSV)')
+    .option('--global-var <value>',
+        'Allows the specification of global variables via the command line, in a key=value format',
+        util.cast.memoizeKeyVal, [])
+    .option('--env-var <value>',
+        'Allows the specification of environment variables via the command line, in a key=value format',
+        util.cast.memoizeKeyVal, [])
+    .option('--export-environment <path>', 'Exports the final environment to a file after completing the run')
+    .option('--export-globals <path>', 'Exports the final globals to a file after completing the run')
+    .option('--export-collection <path>', 'Exports the executed collection to a file after completing the run')
+    .option('--postman-api-key <apiKey>', 'API Key used to load the resources from the Postman API')
+    .option('--bail [modifiers]',
+        'Specify whether or not to gracefully stop a collection run on encountering an error' +
+        ' and whether to end the run with an error based on the optional modifier', util.cast.csvParse)
+    .option('--ignore-redirects', 'Prevents Newman from automatically following 3XX redirect responses')
+    .option('-x , --suppress-exit-code', 'Specify whether or not to override the default exit code for the current run')
+    .option('--silent', 'Prevents Newman from showing output to CLI')
+    .option('--disable-unicode', 'Forces Unicode compliant symbols to be replaced by their plain text equivalents')
+    .option('--color <value>', 'Enable/Disable colored output (auto|on|off)', util.cast.colorOptions, 'auto')
+    .option('--delay-request [n]', 'Specify the extent of delay between requests (milliseconds)', util.cast.integer, 0)
+    .option('--timeout [n]', 'Specify a timeout for collection run (milliseconds)', util.cast.integer, 0)
+    .option('--timeout-request [n]', 'Specify a timeout for requests (milliseconds)', util.cast.integer, 0)
+    .option('--timeout-script [n]', 'Specify a timeout for scripts (milliseconds)', util.cast.integer, 0)
+    .option('--working-dir <path>', 'Specify the path to the working directory')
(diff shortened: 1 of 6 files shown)


raghavbhatia336:
I am Raghav Bhatia, undergrad student from BITS Pialni. I want to apply for GSOC 2021 with Postman. I saw [Idea list](https://docs.google.com/document/d/1F2RWY7wmexsv5KOPsH7T_XiITt8gDK1PPPoaCFf5Ua8/edit#) and one of the ideas is to create `newman request`. 

Goal is `Replacing curl with newman request in a curl command should run the send the request using Newman.`
 
While researching about the topic, I feel one of the best solutions would be to get all `curl options` from the command and and create a Postman collection through [curl-to-postmanv2](https://github.com/postmanlabs/curl-to-postman).

@codenirvana , @shamasis  , @DannyDainton , what  do you think about this idea?

Here is how i plan to implement it:
1. Add a new command `newman request` : https://github.com/raghavbhatia336/newman/blob/feature/single-request/bin/newman.js#L89

2. Newman request will invoke a `newman.request()` function with curl options and newman run options. (To be done)

3. It will pass curl to `curl-to-postmanv2` convert function, and create a Postman collection  with single item (To be moved to `configLoaders`): https://github.com/raghavbhatia336/newman/blob/feature/single-request/lib/run/index.js#L125

4. Use created postman collection and newman options to run the request using `postman-runtime`:  https://github.com/raghavbhatia336/newman/blob/feature/single-request/lib/run/index.js#L179

Let me know if this approach looks good to you. In this PR I have done a small POC for the same.

shamasis:
One last point. 

Getting the parameter match for curl is the easiest part the hardest parts that needs a lot of care is:

1. Making the CLI reporter aware of single request calls in an architecturally neat way
2. Making single request CLI reporter output provide more output and in more helpful details based on the collection / commands
3. Testing and neatly documenting everything

raghavbhatia336:
@shamasis 
Thank you for your feedback! I am glad that you liked this approach. :)

I agree that for `newman request` we can prefix the non-curl newman options in case there is collision with curl options.

I was planning that for default reporters like CLI and also for custom reporters, the information about single request can be made available in the collection's run options, for example, `options.singleRequest` (option name isn't finalized) which are passed to all reporters. I think we cannot add this in reporterOptions because in my understanding, reporterOptions have to be defined by the creator of the Custom Reporter. Am I right?

Although most of the curl options are supported in curl-to-postmanv2, but in case if we find something missing, I'll definitely add it to the library. Also, I will try to have close to 100% test coverage and will follow the documentation guidelines. 

Also, for point 2, since all reporters including CLI reporter will be aware of the newman request/run command, we can modify the output of the reporter based on it. And specifically for default CLI reporter, we can discuss what all extra data we need to report.

shamasis:
I agree with you on all counts. 

I have a proposal to speed things up. 

1. We forget about reporters for single request for now. 

2. We assume that there's no reporter and directly craft a CLI output that has same aesthetics as the CLI reporter.

3. We write the CLI output in such a way that the code can be neatly extracted out to form a new kind of reporter down the line. The way to do this point is to make newman object emit events for single requests (the ones that runtime triggers + any extra if we need) and write CLI output code only off of the events.

This way, we don't need to bother about reporters while we build first cut and then we can add support for new type of reporters or new interceptor events in current reporters. Or we may not even build reporter support for 1st cut until we see a legitimate use case for single request reporter.

raghavbhatia336:
@shamasis 
In my latest commit, I have added the following curl commands: 

- [X] -A, --user-agent 
- [X] -d, --data i
- [X] --data-urlencode
- [ ] --data-binary
- [ ] --data-raw
- [ ] --data-asci
- [X] -F, --form (can be used multiple times)
- [X] -G, --get
- [X] -H, --header (can be used multiple times)
- [X] -X, --request
- [X] -I, --head
- [ ] -T, --file-upload 
- [ ] --url
- [ ] --basic
- [ ] -u, --user

Now, the following example commands are working, with all reporters

```
newman request -X GET http://www.google.com -r cli,json

newman request -X POST  https://postman-echo.com/post -H 'Content-Type: text/plain'  -r progress

newman request -X POST -F 'username=davidwalsh' -F 'password=something' https://postman-echo.com/post

newman request -X POST  https://postman-echo.com/post -d '{"hello":"world"}' -H 'Content-Type: text/json' -H 'Scheme: https' --user-agent 'mobile app' -r cli,progress

newman request -I https://postman-echo.com/post

newman request -G https://google.com

```

shamasis:
This is neat so far. I have not seen newer code or tried it out yet. Let's have the Postman GSOC team to do the actual mentoring and review.

raghavbhatia336:
@shamasis 
Thank you for your feedback. As you suggested, I will remove all the newman run options for now, including reporters. Then I would make the CLI output for single request aesthetically similar to the default CLI reporter. This would certainly speed things up and make it easy to test new changes. I also like your idea for emitting special events for single requests.

shamasis:
Yep. But get your ideas and directions coordinated with your mentor from GSOC team. I'm merely passing by and throwing suggestions where I can when I can.

raghavbhatia336:
@shamasis 

> This is neat so far. I have not seen newer code or tried it out yet. Let's have the Postman GSOC team to do the actual mentoring and review.

I have already sent an email to gsoc@postman.com, but haven't heard from the team yet. Could you please connect me with someone from the GSOC review and mentoring team? That would be really helpful.

shamasis:
There's a community forum you can communicate in - https://community.postman.com/c/open-technology/gsoc/42

raghavbhatia336:
@shamasis 
In my latest commit, I have 
1. created a new function, `newman.request()`, aesthetically similar to `newman.run()`
2. moved curl-to-postman conversion to `configLoaders()` in options.js. 
3. removed all other run options (except reporters) from the newman request for now, to make it easier to test. 

Next, I plan to further change the request function to include the new CLI output for single-request. 
Also, I am thinking of adding new special events for single-request to the emitter. 

Looking forward to your feedback. 
Also, I have still not heard from the GSOC team and don't know about my mentor. So can you please add the concerned team as a reviewer to this PR, so that I can get further feedback from them about my ideas and direction.

PR #2658, opened 2021-03-12
Change the sequence to use object destructuring

aayushmau5:
As I was going through the codebase, I came across this comment
```
    // @todo: ES6: Change the sequence below to use object destructuring when Node v4 support is dropped
   joinPath = nodePath.join,
   parsePath = nodePath.parse,
   resolvePath = nodePath.resolve,
```
As `Node v4` support has been dropped, I think it's safe to use Object destructing instead.

### DIFF
--- a/lib/run/export-file.js
+++ b/lib/run/export-file.js
@@ -5,10 +5,7 @@ var fs = require('fs'),
     async = require('async'),
     mkdirp = require('mkdirp'),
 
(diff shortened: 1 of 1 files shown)


shamasis:
Can you check why tests are failing?

aayushmau5:
Can it be due to the removed lines? But they are still being tested, so not really sure why this is happening. I will look further.
Screenshot_2021-03-15_12 41 02

PR #2664, opened 2021-03-15
Feature: Ignore iterations from start.

### LINKED ISSUE #2168: [feature request] new key to pass iteration data from specified step
I suggest to add ability to skip secified number of iterations from data file

Expected behaviour: I execute command 

    newman run -e env.json -d data.json test.json -n 1

later I need to execute the same test and skip passed iteration:
 the key --skip would provide ability to start test from specified step:

    newman run -e env.json -d data.json test.json --skip 1

adityaofficial10:
Fixes #2168 

**What does this PR do?**
This PR is in response to the feature request received under the above mentioned issue.
Functionality to skip iterations in case of csv files has been added.

- An option --skip-iteration has been added.
- User can skip some specified no. of iterations from the start.
- There is a lower bound and an upper bound on the no. of iterations skipped.

P.S. Tests are coming soon.

### DIFF
--- a/bin/newman.js
+++ b/bin/newman.js
@@ -34,6 +34,8 @@ program
     .option('--env-var <value>',
         'Allows the specification of environment variables via the command line, in a key=value format',
         util.cast.memoizeKeyVal, [])
+    .option('-s,--skip-iteration <value>',
+        'Option to specify the no. of iterations that must be skipped in the data file from the start.')
     .option('--export-environment <path>', 'Exports the final environment to a file after completing the run')
     .option('--export-globals <path>', 'Exports the final globals to a file after completing the run')
     .option('--export-collection <path>', 'Exports the executed collection to a file after completing the run')
--- a/bin/newman.js
+++ b/bin/newman.js
@@ -34,8 +34,8 @@ program
     .option('--env-var <value>',
         'Allows the specification of environment variables via the command line, in a key=value format',
         util.cast.memoizeKeyVal, [])
-    .option('-s,--skip-iteration <value>',
-        'Option to specify the no. of iterations that must be skipped in the data file from the start.')
+    .option('--ignore-iteration <n>',
+        'Option to specify the no. of iterations that must be ignored in the data file from the start.')
     .option('--export-environment <path>', 'Exports the final environment to a file after completing the run')
     .option('--export-globals <path>', 'Exports the final globals to a file after completing the run')
     .option('--export-collection <path>', 'Exports the executed collection to a file after completing the run')
--- a/lib/run/index.js
+++ b/lib/run/index.js
@@ -96,7 +96,8 @@ module.exports = function (options, callback) {
     var emitter = new EventEmitter(), // @todo: create a new inherited constructor
         runner = new runtime.Runner(),
         stopOnFailure,
-        entrypoint;
+        entrypoint,
+        iterationDataLength;
 
     // get the configuration from various sources
     getOptions(options, function (err, options) {
@@ -149,6 +150,16 @@ module.exports = function (options, callback) {
         (typeof options.bail !== 'undefined' &&
             (options.bail === true || (_.isObject(options.bail) && options.bail.failure))) ?
             stopOnFailure = true : stopOnFailure = false;
+
+        if (options.skipIteration) {
+            iterationDataLength = options.iterationData.length;
+            let n = options.skipIteration;
+
+            n = n > 0 && n < iterationDataLength ? n : 0;
+            if (n) {
+                options.iterationData = options.iterationData.slice(n - 1, iterationDataLength - 1);
+            }
+        }
 
         // store summary object and other relevant information inside the emitter
         emitter.summary = new RunSummary(emitter, options);
--- a/lib/run/index.js
+++ b/lib/run/index.js
@@ -97,6 +97,7 @@ module.exports = function (options, callback) {
         runner = new runtime.Runner(),
         stopOnFailure,
         entrypoint,
+        ignoreIteration,
         iterationDataLength;
 
     // get the configuration from various sources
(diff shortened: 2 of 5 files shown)


bin/newman.js
```diff
@@ -34,6 +34,8 @@ program
     .option('--env-var <value>',
         'Allows the specification of environment variables via the command line, in a key=value format',
         util.cast.memoizeKeyVal, [])
+    .option('-s,--skip-iteration <value>',
```
shamasis, line 37: Please do not reserve a single character parameter for a tertiary feature. The single characters should be kept free for major top level features.

Also, could you check the pattern of parameters for iterations and use something along those lines for the double word parameter?

lib/run/index.js
```diff
@@ -150,6 +151,16 @@ module.exports = function (options, callback) {
             (options.bail === true || (_.isObject(options.bail) && options.bail.failure))) ?
             stopOnFailure = true : stopOnFailure = false;
 
+        if (options.skipIteration) {
```
shamasis, line 154: The code flow is weird here. This block will proceed for a negative value. 

The code path complexity for such a tiny feature is unnecessarily high. Please attempt to streamline it.

- first validate
- then arrive at the range
- engage slicing on valid range

shamasis (changes requested):
This is a great idea. ❤️❤️

I have left change recommendation from the code. But more than that I'm not quite sure this is the right name for the solution.

Skip iterations as a term does not match the behaviour of the code. What am I skipping? Where am I skipping from? How many does it skip? All resulting in need to cross refer documentation before usage.

Consider taking input for iteration ignoring from start point reflecting in the parameter name itself.

- [ ] Need tests
- [ ] Need parameter name change
- [ ] Need not to use single character option
- [ ] Need documentation in README
- [ ] Need some form of CLI feedback that iteration was skipped 
- [ ] Code flow itself needs to be simpler and more organised (even for the tiny feature.)

adityaofficial10:
@shamasis 
It will be more meaningful when I complete my patch.
This is just to get it started.
Tests are being written and will be integrated in the next commit.
Thanks for your valuable feedback.
I'll make the required changes.

adityaofficial10:
@shamasis 
i've made the required changes.
How about this?
Screenshot 2021-03-16 at 10 12 40 PM

shamasis:
At this point you should check in and reconcile these with your GSOC mentor for final approvals.

PS: I would rather communicate on public forum for open code. :-)

adityaofficial10:
> At this point you should check in and reconcile these with your GSOC mentor for final approvals.
> 
> PS: I would rather communicate on public forum for open code. :-)

Actually you reviewed it right.
So I thought if I can ask you this?

adityaofficial10:
@coditva @codenirvana 
I've made the required changes as suggested by @shamasis .
Kindly check if it looks fine now..

bin/newman.js
```diff
@@ -34,6 +34,8 @@ program
     .option('--env-var <value>',
         'Allows the specification of environment variables via the command line, in a key=value format',
         util.cast.memoizeKeyVal, [])
+    .option('--ignore-iteration <n>',
```
coditva, line 37: Not sure if `ignore-iteration` is the correct option name. It feels like this should ignore the given iteration number. ​😅

lib/run/index.js
```diff
@@ -96,7 +96,9 @@ module.exports = function (options, callback) {
     var emitter = new EventEmitter(), // @todo: create a new inherited constructor
         runner = new runtime.Runner(),
         stopOnFailure,
-        entrypoint;
+        entrypoint,
+        ignoreIteration,
+        iterationDataLength;
```
coditva, line 101: We can limit the scope of these variables to function on line 102.

lib/run/index.js
```diff
@@ -150,6 +152,11 @@ module.exports = function (options, callback) {
             (options.bail === true || (_.isObject(options.bail) && options.bail.failure))) ?
             stopOnFailure = true : stopOnFailure = false;
 
+        ignoreIteration = options.ignoreIteration;
+        if (options.iterationData) { iterationDataLength = options.iterationData.length; }
+        if (iterationDataLength && ignoreIteration > 0 && ignoreIteration < iterationDataLength) {
+            options.iterationData = options.iterationData.slice(ignoreIteration, iterationDataLength);
+        }
```
coditva, line 159: This block seems to be adding complexity even if `options.ignoreIteration` is not set. Let's prevent this.

test/cli/run-options.test.js
```diff
@@ -125,4 +125,15 @@ describe('CLI run options', function () {
...
+    describe('Option --ignore-iteration ', function () {
+        it('should work correctly when specified', function (done) {
+            // eslint-disable-next-line max-len
+            exec('node ./bin/newman.js run test/integration/steph/steph.postman_collection.json -d test/integration/steph/steph.postman_data.json --ignore-iteration 1', function (code, stdout, stderr) {
+                expect(code, 'should have zero exit code').to.equal(0);
+                expect(stderr).to.be.empty;
+                done();
+            });
+        });
+    });
```
coditva, line 138: This test seems to be checking that the run was successful. But it does not check if the option is working correctly.

test/library/run-options.test.js
```diff
@@ -239,4 +241,29 @@ describe('Newman run options', function () {
             });
         });
     });
+
+    describe('Option --ignore-iteration ', function () {
+        var iterationProperty = 'run.stats.iterations.total',
+            collectionRunPath = path.join(__dirname, '..', '..', 'out', 'iteration-count-test.json');
```
coditva, line 247: You can simplify this:
```suggestion
            collectionRunPath = path.join(__dirname, '../../out/iteration-count-test.json');
```

coditva, line 247: We can limit the scope of these variables to the `it` block.

test/unit/options.test.js
```diff
@@ -97,4 +97,27 @@ describe('options', function () {
...
+            options({
+                iterationData: './test/fixtures/run/spaces/data.json',
+                ignoreiteration: 1
+            }, function (err, result) {
+                expect(err).to.be.null;
+                expect(result.ignoreiteration).to.equal(1);
+                done();
+            });
+        });
+    });
```
coditva, line 122: We should also check cases where ignoreIteration has invalid values like `-1`, `> length of iteration data`, `Infinity`, etc.

coditva (changes requested):
@adityaofficial10 Added review comments inline.

One thought, however:
We're only thinking about one use-case where users want to skip iterations at the beginning. Can we make this more generic by allowing the users to pass a range?

Thoughts? @codenirvana

adityaofficial10:
@coditva 
Thanks for your valuable feedback.
I'll make the changes shortly..
As for the suggestion,  I was looking forward to it and if permitted, I can upgrade this...

shamasis:

All the above seems to be risky to implement even when it does what it is supposed to. These should be done at runtime level. Instead of manipulating data file, runtime itself should start and end on specific iterations. It would also help it make portable across CLI and App

PR #2665, opened 2021-03-15
add file extension based loading

sambhavjai:
Going through the code base there was a todo marked at the `lib\run\options.js` in `iterationData` function to change the loading of iteration data to extension based loading. Following changes have been made in the file.

- remove `async.waterfall` function
- getting the file extension from the file location
- parsing the file according to the extension
- remove the import of `async` as it was no longer required

### DIFF
--- a/lib/run/options.js
+++ b/lib/run/options.js
@@ -272,7 +272,7 @@ var _ = require('lodash'),
                     return callback(new Error(ITERATION_DATA_LOAD_ERROR_MESSAGE + `\n  ${err.message || err}`));
                 }
 
-                // Try loading as a JSON, fall-back to CSV. @todo: switch to file extension based loading.
+                // Try loading as a JSON, fall-back to CSV.
                 async.waterfall([
                     (cb) => {
                         try {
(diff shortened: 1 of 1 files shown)


shamasis:
@sambhavjai neat attempt. But I think it's too late. We left this todo unattended for very long. As such, I'm sure we now have many usage setup where users are not bothered about file extension and is even feeding newman with iteration data from various file extensions (and even from files with no extension.)

At this moment this would break user behaviour and thus reducing the incentive to implement this todo.

I would rather undo all the changes and remove the todo as the PR. 

Sorry dear. 

@codenirvana Cc-ing you if you have anything different in your mind.

sambhavjai:
@shamasis thanks for the review. I will wait for @codenirvana to comment on this if he wants something else to be done with this todo otherwise I will remove the todo as the PR.

sambhavjai:
@shamasis I have undone the changes and removed the todo. I guess this is now ready to merge.

shamasis (approved):

shamasis (approved):

Issue #2666, opened 2021-03-15
Enhacement: Descriptive white space reporter error message


elit-altum:
1. Newman Version (can be found via `newman -v`): 5.2.2
 2. OS details (type, version, and architecture): Windows 10 x64
 3. Are you using Newman as a library, or via the CLI? CLI
 3. Did you encounter this recently, or has this bug always been there: Not a bug
 4. Expected behaviour: More descriptive error message regarding not using white spaces between reporter names.
 5. Command / script used to run Newman: ```newman run examples/sample-collection.json -r cli, json```

Though the documentation clearly mentions that while mentioning multiple reporter names the user should not use spaces after the commas, this can be looked over or forgotten while running newman. The error thrown in such a case does not point out this issue, which would enable quicker debugging. 
I would like to propose to add a line in the error statement to consider removing the white space before reporter names.

**BEFORE**
![image](https://user-images.githubusercontent.com/41413622/111199831-1ae3ac00-85e7-11eb-94b1-b59b8c9194fd.png)

**AFTER**
![image](https://user-images.githubusercontent.com/41413622/111199747-02739180-85e7-11eb-94ab-dd6bc36a03bf.png)

I have a PR ready for this, I would like to discuss the exact wordings of the error messages before publishing the PR 🚀

elit-altum:
@shamasis Please have a look!

shamasis:
Yeah definitely - this needs fixing. But is there a way we can completely get rid of this error scenario?

If it is a parsing error, can we not prune blank reporter options before passing the options further down the line in code?

I would rather attempt at fixing the root cause than suppress the symptom. Only and only if the root cause resolution is a massive endeavour is when I would go the message route for this case.

elit-altum:
Absolutely @shamasis, I tried to find the location where the parsing is happening but I wasn't able to locate it, is it a ```utils``` function?
Could you please guide me in the right direction?

shamasis:
I'm afraid that would be very difficult for me at the moment.

Try putting a breakpoint on the error and observe the stack trace. 

Basic CLI parsing is done using an external library and then a bit of organising is done on top of it. Most of the code around that is in the cli bin js file.

johnmusk:
@shamasis Why this issue is still open I think @elit-altum solve the issue or not

shamasis:
Yeah. Waiting for @codenirvana to review once. He's a bit stuck on some other releases. Let me see if I can bump it up.

PR #2668, opened 2021-03-16
fix: sanitize whitespaces and commas before reporter names

elit-altum:
Potential fix : #2666 

It seems that commander.js does not have inbuilt support for parsing any arguments after white spaces which leads to the error. I have manually processed all the reporter options and sanitized the reporter names from there. 
The program now works for extreme cases as well [CHECK SCREENSHOTS].

I would like some help on the coding practices like variable naming, file splitting etc. so my code complies well with the codebase.

The command and the reporters array is as follows: 

CASE 1 : Original
image

CASE 2: With Whitespace
image

CASE 3: Extreme Case 1
image

CASE 4: Extreme Case 2
image

@shamasis @codenirvana 
Please have a look!


adityaofficial10:
@elit-altum  @shamasis 
I have a simpler solution which can save us a lot of time here.
We can discuss it if you want.

elit-altum:
Yeah sure @adityaofficial10 !

adityaofficial10:
@elit-altum  @shamasis 
Why not use a variadic option for reporters?
That would eliminate the need of commas and make the syntax easy for users.

bin/newman.js
```diff
@@ -62,11 +62,64 @@ program
...
+                if (sanitizedReporterName.slice(-1) === ',') {
+                    sanitizedReporterName = sanitizedReporterName.split(',').join('');
+                }
+
+                reporters[reporterIndex] = sanitizedReporterName;
+            });
+
+            command.reporters = reporters;
+        }
         let options = util.commanderToObject(command),
```
shamasis, line 65: What does this util do? Asking since I want to check if the pattern is to put put reusable CLI processing in util. Which is also not a good pattern, but still ... asking.

shamasis:
Variadic parameter is a breaking change. Is it not? And "optional parameter" + "variadic nature" does not play well together.

adityaofficial10:
> Variadic parameter is a breaking change. Is it not? And "optional parameter" + "variadic nature" does not play well together.

Yeah thats actually true but I just thought that it would make the syntax simpler.
This way, we'll have to integrate a lot of manual parsing.

bin/newman.js
elit-altum: @shamasis This is what the method comments say
```js
/**
     * Extract selected options in the provided command.
     * Omits commander private variables and other objects.
     *
     * @param {Object} command - Commander.Command Instance
     * @returns {Object} - Extracted options from command
     */
```

I think CLI processing is in util only.

shamasis:
@elit-altum the general direction is great. But something's off with this. I'm unable to pin it yet. There's a feeling that the codeflow can be simpler and that there's a simpler way.

Let me once re-look at it on my next available slot. 

One thought, what if we don't touch the CLI processing, but simply ignore blank reporter names at the place in newman where reporters are registered?

elit-altum:
@shamasis Please suggest next steps with this, should I start working on the tests? or do we need some refactoring here?

elit-altum:
> 
> 
> @elit-altum the general direction is great. But something's off with this. I'm unable to pin it yet. There's a feeling that the codeflow can be simpler and that there's a simpler way.
> 
> Let me once re-look at it on my next available slot.
> 
> One thought, what if we don't touch the CLI processing, but simply ignore blank reporter names at the place in newman where reporters are registered?

Sure @shamasis I will also dig deeper into this. 

The problem is that commander itself stops parsing options after " ". The parsed reporter names sent to us by commander for ```-r cli, json``` is ```['cli', ' ']```. There is no reference of anything after the space. That is why I had to manually process the args.

adityaofficial10:
@elit-altum 
You're right.
I also noticed the same thing.
I guess we'll have to dig it deep to find the solution.

bin/newman.js
```diff
@@ -62,11 +62,64 @@ program
     .option('--export-cookie-jar <path>', 'Exports the cookie jar to a file after completing the run')
     .option('--verbose', 'Show detailed information of collection run and each request sent')
     .action((collection, command) => {
+        // Process reporter arguments explicitly
+        const rawArgs = command.parent.args,
+            isReporterPresent = rawArgs.includes('-r') || rawArgs.includes('--reporter');
+
+        if (isReporterPresent) {
+            const reporters = ['cli'];
+            let reporterIndex = rawArgs.indexOf('-r'),
```
abhijit-hota, line 71: Instead of reassigning `reporterIndex` again (in case of `--reporter`), I think we can do this:
```js
const reporterIndex = rawArgs.indexOf('-r') !== -1 ? rawArgs.indexOf('-r') : rawArgs.indexOf('--reporter')
```

Or, compromising readability a bit, the following:

```js
const reporterIndex = Math.max(rawArgs.indexOf('-r'), rawArgs.indexOf('--reporter'))
```

P.S.
A question: What happens in the case of both `-r` and `--reporter` being passed?

elit-altum: Thanks for the suggestion, but statement 1 gives us a es-lint error and no compromises can be made on readability.
Second, this PR is redundant as we have found a better way around this so I'd be closing this.
😄

abhijit-hota: Yeah. Just saw that we're going with variadic args, which imo, is a better choice.  🎉

PR #2673, opened 2021-03-17
add: dedicated method for parsing reporters

elit-altum:
@shamasis @codenirvana  

This is an alternate way of handling #2668  using variadic parameters. I think this is a cleaner solution and achieves the same results.

Please have a look!

### DIFF
--- a/bin/newman.js
+++ b/bin/newman.js
@@ -22,7 +22,9 @@ program
     .usage('<collection> [options]')
     .option('-e, --environment <path>', 'Specify a URL or path to a Postman Environment')
     .option('-g, --globals <path>', 'Specify a URL or path to a file containing Postman Globals')
-    .option('-r, --reporters [reporters...]', 'Specify the reporters to use for this run', util.cast.reporterParse, ['cli'])
+    .option('-r, --reporters [reporters...]',
+        'Specify the reporters to use for this run',
+        util.cast.reporterParse, ['cli'])
     .option('-n, --iteration-count <n>', 'Define the number of iterations to run', util.cast.integer)
     .option('-d, --iteration-data <path>', 'Specify a data file to use for iterations (either JSON or CSV)')
     .option('--folder <path>',
--- a/bin/newman.js
+++ b/bin/newman.js
@@ -22,7 +22,7 @@ program
     .usage('<collection> [options]')
     .option('-e, --environment <path>', 'Specify a URL or path to a Postman Environment')
     .option('-g, --globals <path>', 'Specify a URL or path to a file containing Postman Globals')
-    .option('-r, --reporters [reporters]', 'Specify the reporters to use for this run', util.cast.csvParse, ['cli'])
+    .option('-r, --reporters [reporters...]', 'Specify the reporters to use for this run', util.cast.reporterParse)
     .option('-n, --iteration-count <n>', 'Define the number of iterations to run', util.cast.integer)
     .option('-d, --iteration-data <path>', 'Specify a data file to use for iterations (either JSON or CSV)')
     .option('--folder <path>',
--- a/bin/newman.js
+++ b/bin/newman.js
@@ -85,9 +85,9 @@ program
         });
     });
 
-// program.addHelpText('after', `
-// To get available options for a command:
-//   newman <command> -h`);
+program.addHelpText('after', `
+To get available options for a command:
+  newman <command> -h`);
 
 // Warn on invalid command and then exits.
 program.on('command:*', (command) => {
--- a/bin/newman.js
+++ b/bin/newman.js
@@ -22,7 +22,7 @@ program
     .usage('<collection> [options]')
     .option('-e, --environment <path>', 'Specify a URL or path to a Postman Environment')
     .option('-g, --globals <path>', 'Specify a URL or path to a file containing Postman Globals')
-    .option('-r, --reporters [reporters...]', 'Specify the reporters to use for this run', util.cast.reporterParse)
+    .option('-r, --reporters [reporters...]', 'Specify the reporters to use for this run', util.cast.reporterParse, ['cli'])
     .option('-n, --iteration-count <n>', 'Define the number of iterations to run', util.cast.integer)
     .option('-d, --iteration-data <path>', 'Specify a data file to use for iterations (either JSON or CSV)')
     .option('--folder <path>',
@@ -85,9 +85,9 @@ program
         });
     });
 
-program.addHelpText('after', `
-To get available options for a command:
-  newman <command> -h`);
+// program.addHelpText('after', `
+// To get available options for a command:
+//   newman <command> -h`);
 
 // Warn on invalid command and then exits.
 program.on('command:*', (command) => {
--- a/bin/util.js
+++ b/bin/util.js
@@ -85,7 +85,6 @@ module.exports = {
          * @returns {String[]} - [item1, item2] - The array representation of the passed parameters.
          */
         reporterParse: (list) => {
-            list.trim();
             if (list.includes(',')) {
                 list.split(',').forEach((name) => {
                     if (name.length) {
--- a/bin/util.js
+++ b/bin/util.js
@@ -1,4 +1,6 @@
-const _ = require('lodash');
+const _ = require('lodash'),
+    reporters = ['cli'];
+
 
 module.exports = {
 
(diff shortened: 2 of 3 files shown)


adityaofficial10:
@elit-altum 
It has been already discussed that variadic method is not an option as that would be a breaking change
Check this out: https://github.com/postmanlabs/newman/pull/2668

elit-altum:
> @elit-altum
> It has been already discussed that variadic method is not an option as that would be a breaking change
> Check this out: #2668

I know that #adityaofficial10, I authored that PR 😅

But this works similar to the previous command and gives the same functionality, which is what we were looking for.

adityaofficial10:
@elit-altum 
@shamasis has already clarified that breaking changes would be avoided.
So this will help you only.

shamasis:
At a first glance @elit-altum this PR looks exactly like what and how we wanted it.

Get the final approval from your GSOC mentor and get this merged? It will be good.

elit-altum:
> At a first glance @elit-altum this PR looks exactly like what and how we wanted it.
> 
> Get the final approval from your GSOC mentor and get this merged? It will be good.

Thank you for the review @shamasis!

@codenirvana Could you please have a look at this so we can merge this?

shamasis:
One thing though ... I haven't tested or checked whether any external usage has changed or not. Please check the tests and also ensure that existing usage doesn't require changes

elit-altum:
@shamasis sure, I would once again test this and add new unit tests for the same!

elit-altum:
@shamasis  @adityaofficial10 
Something unrelated, is this error coming for your system as well when you run newman: 

```
program.addHelpText('after', `
        ^

TypeError: program.addHelpText is not a function
    at Object.<anonymous> (D:\Safehouse\Code\Open Source\newman\bin\newman.js:88:9)
    at Module._compile (internal/modules/cjs/loader.js:1063:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1092:10)
    at Module.load (internal/modules/cjs/loader.js:928:32)
    at Function.Module._load (internal/modules/cjs/loader.js:769:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:72:12)
    at internal/main/run_main_module.js:17:47
```

I believe this has started coming after https://github.com/postmanlabs/newman/commit/5239507ead88fd519a80465e42d26e6f748ca1b8

shamasis:
Try cleaning your node modules folder and reinstall. But, no, I've not faced this till yesterday.

Move this unrelated conversation to a separate issue or to the community. 😊

adityaofficial10:
@elit-altum 
Bro try to read the logs of Travis CI.
That would help.
From my view, I think there's an ESLint error somewhere related to maxlen. 
Also one of the unit tests is failing.
Just check if the change you made to reporter option has been updated there.
This should help.😄

abhijit-hota:
Why do we need to split according to the commas if we're using variadic parameters?

abhijit-hota:
> @elit-altum
> Bro try to read the logs of Travis CI.
> That would help.
> From my view, I think there's an ESLint error somewhere related to maxlen.
> Also one of the unit tests is failing.
> Just check if the change you made to reporter option has been updated there.
> This should help.😄

I think that can be solved by introducing newlines like this:

```js
.option('-r, --reporters [reporters...]', 
        'Specify the reporters to use for this run',
        util.cast.reporterParse, ['cli'])
```

bin/newman.js
```diff
@@ -22,7 +22,7 @@ program
     .usage('<collection> [options]')
     .option('-e, --environment <path>', 'Specify a URL or path to a Postman Environment')
     .option('-g, --globals <path>', 'Specify a URL or path to a file containing Postman Globals')
-    .option('-r, --reporters [reporters]', 'Specify the reporters to use for this run', util.cast.csvParse, ['cli'])
+    .option('-r, --reporters [reporters...]', 'Specify the reporters to use for this run', util.cast.reporterParse, ['cli'])
```
abhijit-hota, line 25: I think the default `['cli']`  option can be removed as we're returning that value from the cast util.

bin/util.js
```diff
@@ -74,6 +76,30 @@ module.exports = {
...
+         * eg. item1,item2
+         *
+         * @param {String} list - The comma separated parameters.
+         * @returns {String[]} - [item1, item2] - The array representation of the passed parameters.
+         */
+        reporterParse: (list) => {
+            list.trim();
+            if (list.includes(',')) {
+                list.split(',').forEach((name) => {
+                    if (name.length) {
```
abhijit-hota, line 91: The `cli` arg gets added twice in this case. Just a minor suggestion.

```diff
- if (name.length) {
+ if (name.length && name !== 'cli') {
```

bin/newman.js
elit-altum: This is present as a default if someone does not provide any arguments after the ```-r``` command.

bin/util.js
elit-altum: Handling the previous edge case, this is important. The reporter does take care of checking duplicates before running it, so this isn't an issue.

bin/newman.js
abhijit-hota: Oh, I see. Thanks for the clarification.

Sharath-Postman:
@elit-altum do create an issue on community.postman.com gsoc category under open technologies. One of the mentors from the newman team will review this

bin/util.js
```diff
@@ -74,6 +76,30 @@ module.exports = {
...
+        /**
+         * Helper to parse variadic parameters for reporters to an array.
+         *
+         * eg. item1,item2
+         *
+         * @param {String} list - The comma separated parameters.
+         * @returns {String[]} - [item1, item2] - The array representation of the passed parameters.
+         */
+        reporterParse: (list) => {
+            list.trim();
```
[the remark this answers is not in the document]
elit-altum: Yes that's true @raghavgarg1257, I missed this, my bad 😅
Actually commander.js is handling the trimming for us so there is no need to use this. The required changes have been made.

Thank You!

raghavgarg1257:
@elit-altum Can you please fix the test to add the code coverage for the code that has been added for this PR?
Code Ref: https://github.com/postmanlabs/newman/blob/develop/test/unit/cli.test.js#L208

elit-altum:
> 
> 
> @elit-altum Can you please fix the test to add the code coverage for the code that has been added for this PR?
> Code Ref: https://github.com/postmanlabs/newman/blob/develop/test/unit/cli.test.js#L208

Sure @raghavgarg1257 are you suggesting something like changing [this line](https://github.com/postmanlabs/newman/blob/24d13f7a8d6e39ef75093fb3efa69cd93daa2179/test/unit/cli.test.js#L224) to something like:
```--reporters json, cli```

and then add a line after [this one](https://github.com/postmanlabs/newman/blob/24d13f7a8d6e39ef75093fb3efa69cd93daa2179/test/unit/cli.test.js#L271) to check for a``` cli``` reporter as well ?

**Comment by raghavgarg1257** [comment]
@elit-altum Yes, but maybe you will have to use other reporters for testing because `cli` has some conflicting config already present there.

**Comment by elit-altum** [comment]
@raghavgarg1257 Added the unit tests!
I have included ```progress``` as the other reporter as there was already a test to ensure ```junit``` is not included.

File: bin/util.js [hunk 4]
```diff
@@ -74,6 +76,29 @@ module.exports = {
             return _.split(list, ',');
```
codenirvana, line 76: We just need to do this:
```suggestion
            return list.split(',').map((item) => { return item.trim(); });
```

codenirvana (changes requested):

bin/util.js
elit-altum: @codenirvana I tried this, but I believe commander for variadic parameters will run this function for each parameter individually i.e. if I give ```-r cli, json```, in the above function ```list``` would first be ```"cli"``` and then ```" json"```, therefore the final reporters parsed would just be ```['json']``` coming from the last ```list``` value.

While if we do not choose variadic parameters, commander will stop parsing after the space between ```cli, json``` and we won't be able to achieve this functionality.

elit-altum: I have reduced the function to this now,
```js
let reporters = [];

/**
         * Helper to parse variadic parameters for reporters to an array.
         *
         * eg. item1,item2
         *
         * @param {String} list - The comma separated parameters.
         * @returns {String[]} - [item1, item2] - The array representation of the passed parameters.
         */
        reporterParse: (list) => {
            list.split(',').forEach((name) => {
                if (name.length) {
                    reporters.push(name);
                }
            });

            return reporters;
        },

```

It's a little bit more concise than before and works the same. Should I commit this change?

codenirvana: What I meant was, the only change required is to trim the `csvParse` values, no commander change required.

Because variadic option will allow input like: `-r cli json` which is a major change in my opinion.
We can introduce this behavior change in v6 for all similar options and deprecate this.

elit-altum: @codenirvana Sorry for the misunderstanding there.

But if I do not use variadic parameters with reporters, commander stops parsing the parameters after the space i.e. if I use ```-r cli, json``` the ```list``` argument in ```csvParse``` will only have the value ```cli,``` nothing else.

Issue #1991, opened 2019-05-03
Newman doesn't appear to be able to download Postman Collections in Node 12


deltamualpha:
1. Newman Version (can be found via `newman -v`): 4.4.1
 2. OS details (type, version, and architecture): macOS 10.14.4; node 12.1.0
 3. Are you using Newman as a library, or via the CLI? **CLI**
 3. Did you encounter this recently, or has this bug always been there: New installation
 4. Expected behaviour: It runs a collection
 5. Command / script used to run Newman: `newman run https://raw.githubusercontent.com/heremaps/postman-collections/master/weather.postman_collection`
 6. Sample collection, and auxiliary files (minus the sensitive details):
 7. Screenshots (if applicable):

Node 12.1.0:
```
$ newman run https://raw.githubusercontent.com/heremaps/postman-collections/master/weather.postman_collection
newman: Newman v4 deprecates support for the v1 collection format
  Use the Postman Native app to export collections in the v2 format

newman

Weather API

→ Severe Weather Alerts
_tls_wrap.js:513
  self._destroySSL();
       ^

TypeError: Cannot read property '_destroySSL' of undefined
    at Timeout.destroySSL [as _onTimeout] (_tls_wrap.js:513:8)
    at listOnTimeout (internal/timers.js:531:17)
    at processTimers (internal/timers.js:475:7)
```

Node 10:
```
$ newman run https://raw.githubusercontent.com/heremaps/postman-collections/master/weather.postman_collection
newman: Newman v4 deprecates support for the v1 collection format
  Use the Postman Native app to export collections in the v2 format

newman

Weather API

→ Severe Weather Alerts
  GET https://weather.cit.api.here.com/weather/1.0/report.json?product=alerts&name=Chicago&app_id={{YOUR_APP_ID}}&app_code={{YOUR_APP_CODE}} [401 Unauthorized, 550B, 156ms]
... etc etc etc
```

Node 12:
```
$ newman run weather.postman_collection
newman: Newman v4 deprecates support for the v1 collection format
  Use the Postman Native app to export collections in the v2 format

newman

Weather API

→ Severe Weather Alerts
  GET https://weather.cit.api.here.com/weather/1.0/report.json?product=alerts&name=Chicago&app_id={{YOUR_APP_ID}}&app_code={{YOUR_APP_CODE}} [401 Unauthorized, 550B, 137ms]
... etc etc etc
```

deltamualpha:
[...]

PR #2679, opened 2021-03-23
chore: update to WHATWG url api

elit-altum:
NodeJS has deprecated the [Legacy URL API](https://nodejs.org/api/deprecations.html#deprecations_dep0116_legacy_url_api) which was being used in ```utils.js``` and recommends moving to the [WHATWG URL API](https://nodejs.org/api/url.html#url_the_whatwg_url_api).

I don't believe this change can fix https://github.com/postmanlabs/newman/issues/1991 but I'll be happy to look into it deeper.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -194,7 +194,7 @@ util = {
                     urlObj = new URL(location);
 
                     (urlObj.hostname === POSTMAN_API_HOST) &&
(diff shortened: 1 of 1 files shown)


shamasis (approved):
Since new URL works v7 onwards and the change is only in fetching data from command line collection and env url, this should not have huge blast radius.

codenirvana (approved):

PR #2688, opened 2021-03-29
refactor: added extension based parsing - lib/run/options.js

elit-altum:
Fixed a ```@todo``` to enable file extension based loading while parsing CSV/JSON in ```lib/run/options.js``` using```path.extname```

@shamasis @codenirvana 
Please review!

### DIFF
--- a/lib/run/options.js
+++ b/lib/run/options.js
@@ -267,24 +267,16 @@ var _ = require('lodash'),
         iterationData: function (location, options, callback) {
             if (_.isArray(location)) { return callback(null, location); }
 
-            util.fetch(location, function (err, data) {
+            util.fetch(location, function (err, data, fileType) {
                 if (err) {
                     return callback(new Error(ITERATION_DATA_LOAD_ERROR_MESSAGE + `\n  ${err.message || err}`));
                 }
 
-                // Try loading as a JSON, fall-back to CSV. @todo: switch to file extension based loading.
+                // Check if the file is JSON else parse as CSV
                 async.waterfall([
                     (cb) => {
-                        try {
+                        if (fileType === '.json') {
                             return cb(null, liquidJSON.parse(data.trim()));
-                        }
-                        catch (e) {
-                            return cb(null, undefined); // e masked to avoid displaying JSON parse errors for CSV files
-                        }
-                    },
-                    (json, cb) => {
-                        if (json) {
-                            return cb(null, json);
                         }
                         // Wasn't JSON
                         parseCsv(data, {
--- a/lib/run/options.js
+++ b/lib/run/options.js
@@ -276,7 +276,12 @@ var _ = require('lodash'),
                 async.waterfall([
                     (cb) => {
                         if (fileType === '.json') {
-                            return cb(null, liquidJSON.parse(data.trim()));
+                            try {
+                                return cb(null, liquidJSON.parse(data.trim()));
+                            }
+                            catch (e) {
+                                return cb(e, undefined);
+                            }
                         }
                         // Wasn't JSON
                         parseCsv(data, {
--- a/lib/util.js
+++ b/lib/util.js
@@ -1,5 +1,6 @@
 var fs = require('fs'),
     url = require('url'),
+    path = require('path'),
 
     _ = require('lodash'),
     chardet = require('chardet'),
@@ -248,7 +249,7 @@ util = {
                     return callback(err);
                 }
 
-                return callback(null, value.toString(util.detectEncoding(value)));
+                return callback(null, value.toString(util.detectEncoding(value)), path.extname(location));
             });
     },
 


lib/run/options.js
```diff
@@ -267,25 +267,17 @@ var _ = require('lodash'),
...
                     return callback(new Error(ITERATION_DATA_LOAD_ERROR_MESSAGE + `\n  ${err.message || err}`));
                 }
 
-                // Try loading as a JSON, fall-back to CSV. @todo: switch to file extension based loading.
+                // Check if the file is JSON else parse as CSV
                 async.waterfall([
                     (cb) => {
-                        try {
+                        if (fileType === '.json') {
                             return cb(null, liquidJSON.parse(data.trim()));
```
raghavbhatia336, line 279: @elit-altum I think here `try...catch` should be there to catch .json parsing errors.

elit-altum: This was skipped as the previous ```try-catch``` code was just for catching the is JSON or not error. Therefore, for an invalid JSON the error message provided by newman was 
```
error: iteration data could not be loaded
 Invalid Closing Quote: found non trimable byte after quote at line 2
```

this was because ```parseCSV``` was handling the JSON error.

elit-altum:
Invalid JSON

```json
{
  "foo": "bar",
  "lorem": "ipsum",
  "country": "India"
  "postman": "newman"
}
```

When this file is used in ```newman``` using ```newman run examples/sample-collection.json -d ./newman/check/check.json```

The error returned is:
```error: iteration data could not be loaded
 Invalid Closing Quote: found non trimable byte after quote at line 2
```

this is because, JSON was being parsed by ```parseCSV``` in the original flow of the programs

Updated fix for giving errors by ```liquidJSON```:

```
node ./bin/newman run examples/sample-collection.json -d ./newman/check/check.json
```

Error returned is :
```
error: iteration data could not be loaded
 Unexpected token 'p' at 5:4
  "postman": "newman"
```

shamasis:
Hi. Apologies but this todo has expired and there's a pending PR merge to address the same. You can find reasoning here - https://github.com/postmanlabs/newman/pull/2665

shamasis:
Merged this to avoid confusion the stray todo was causing. Need a retrospective sign-off from @codenirvana

elit-altum:
> 
> 
> Invalid JSON
> 
> ```json
> {
>   "foo": "bar",
>   "lorem": "ipsum",
>   "country": "India"
>   "postman": "newman"
> }
> ```
> 
> When this file is used in `newman` using `newman run examples/sample-collection.json -d ./newman/check/check.json`
> 
> The error returned is:
> 
> ```
>  Invalid Closing Quote: found non trimable byte after quote at line 2
> ```
> 
> this is because, JSON was being parsed by `parseCSV` in the original flow of the programs
> 
> Updated fix for giving errors by `liquidJSON`:
> 
> ```
> node ./bin/newman run examples/sample-collection.json -d ./newman/check/check.json
> ```
> 
> Error returned is :
> 
> ```
> error: iteration data could not be loaded
>  Unexpected token 'p' at 5:4
>   "postman": "newman"
> ```

Sure @shamasis !
But what about this issue? The current invalid JSON data is being parsed by ```parseCSV``` and the errors returned by it are not descriptive. Do we want to fix that or let it be ?

shamasis:
Yes to fix this the PR has to be a lot more creative. Whatever you do will lead to some error. If you change order of parsing, then invalid CSV files will see JSON error.

The long way out would be:

1. Use current parsing logic
2. On error, keep all error messages stored
3. If both CSV and JSON parsing fails, then use extension to show the appropriate error

If @codenirvana agrees, then maybe a separate PR will be good for this. But I will definitely wait for @codenirvana to chime in.

elit-altum:
Sure @shamasis , I have made the changes here for handling errors in all the 3 situations. 

If @codenirvana thinks they are okay I'll be happy to create a new PR for fixing those!

elit-altum:
> 
> 
> Yes to fix this the PR has to be a lot more creative. Whatever you do will lead to some error. If you change order of parsing, then invalid CSV files will see JSON error.
> 
> The long way out would be:
> 
>     1. Use current parsing logic
> 
>     2. On error, keep all error messages stored
> 
>     3. If both CSV and JSON parsing fails, then use extension to show the appropriate error
> 
>     4. If extension is neither JSON or CSV ... then 🤷‍♂️ send an API call to UN council to send humanitarian aid. (Jokes aside, the last case then can show both errors by letting users know we tried parsing as both formats?)
> 
> 
> If @codenirvana agrees, then maybe a separate PR will be good for this. But I will definitely wait for @codenirvana to chime in.

@codenirvana @coditva Could you please have a look at this so I can draft a new PR for fixing this!
Thank You!

PR #2694, opened 2021-04-05
chore: ignore yarn.lock

aman-jaglan:
Ignore `yarn.lock` file to prevent getting commited accidently.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,9 @@
 # - Do not allow installed node modules to be committed. Doing `npm install -d` will bring them in root or other places.
 node_modules
 
+# Ignore lock files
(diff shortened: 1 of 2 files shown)


elit-altum (commented):
Hey @aman-jaglan I don't think there is a need for ```yarn.lock``` as the contributing guidelines clearly specify to use ```npm```. [REFER [HERE](https://github.com/postmanlabs/newman/blob/develop/.github/CONTRIBUTING.md#getting-started-quick)]

aman-jaglan:
Yes we don't use `yarn` hence there is no harm in ignoring it, I have seen contributors commiting this file accidently.

Issue #2725, opened 2021-04-28
[Feature Request] Include more request information under the `--verbose` flag


DannyDainton:
Currently using the `--verbose` flag will only show the detailed timings for each request:

![image](https://user-images.githubusercontent.com/17089546/116439733-c408fd80-a847-11eb-8adf-c0f3ece9e493.png)

The documentation for the `--verbose` flag seems to suggest that this flag would expose _more_ details and not just the timings.

> --verbose
Show detailed information of collection run and each request sent.

Could this flag be extended to show additional details about the request/response headers or the request/response bodies to allow uses to easily debug issues during the run?

Additional details could be exposed by the user by using `NODE_DEBUG=http newman run collection.json` command but it would be good to give a reduced version of this under the handy `--verbose` flag.

adityaofficial10:
Hey @DannyDainton !
Hope you're doing well.
This time I'm not troubling you.😂
What I think is we can easily render information like response headers and redirect information here.
What you say?

DannyDainton:
Hey @adityaofficial10 👋🏻 

It's not really up to me what gets added or how it might look in the console output - It's just a placeholder for a conversation for expanding the `--verbose` flag, to show more than only the detailed timings 😄 

If you feel you have something in mind, you could create a PR for the team to take a look at and see if that's the direction we would like to take.

raghavbhatia336:
@DannyDainton I have added response header and request headers in verbose, let me know if you would like to add other request details like, for example, request body.
Is this something we're looking for?

![image](https://user-images.githubusercontent.com/38736520/117043875-65ea8780-ad2b-11eb-98f2-2d8ecb15554d.png)

noahlz:
I'm looking for an extra-verbose output... currently my team is fighting against mysterious "parse error" test failures from newman that imply network or even OS-level issues (we're running newman in docker).

curl verbosity looks like this:

```
$ curl -v -v -v https://google.com
* Rebuilt URL to: https://google.com/
* Hostname was NOT found in DNS cache
*   Trying 142.250.65.78...
* Connected to google.com (142.250.65.78) port 443 (#0)
* successfully set certificate verify locations:
*   CAfile: none
  CApath: /etc/ssl/certs
* SSLv3, TLS handshake, Client hello (1):
* SSLv3, TLS handshake, Server hello (2):
```

We need something close to this level of verbosity from newman output to help us track down non-json content level errors.

noahlz:
[...]

PR #2730, opened 2021-05-04
Add request and response header in verbose

raghavbhatia336:
[Issue](https://github.com/postmanlabs/newman/issues/2725)

Added request and response headers to --verbose

image

@DannyDainton is this something what we're looking for?


elit-altum:
Can we look at integrating somethings from here as well? https://www.npmjs.com/package/newman-reporter-cli-response
I think it would help.

Also curious about what you think about the formatting of the headers, did you have any specific reason for logging the exact json? 🤔

Issue #2752, opened 2021-05-31
Newman summary JSON having incorrect values


khushyadav:
If a request contains pm.sendRequest() in its test then, the summary object created post the execution of collection using newman.run yields following incorrect values:

1. Under summary.run.executions we would have 2 object with identical values . One corresponding to request and second corresponding to pm.sendRequest.
2. The url of parent request will also be incorrectly replaced with the url of pm.sendrequest() under summary.run.executions[index].request.url
Also newman report created also shows incorrect endpoints for this kind of case.

Information Required:
 1. Newman Version : **5.0.0**
 2. OS details: **Windows 10** 
 3. Are you using Newman as a library, or via the CLI: **As library**
 3. Did you encounter this recently, or has this bug always been there: **Always been there**
 5. Command / script used to run Newman: 
 newman.run({
          collection: collection,
          globals: global,
          environment: env,
          iterationData: data,
          reporters: reporters,
          reporter: {
            htmlextra: {
              export: report,
              template: templateFile,
              logs: true
            }
          },
          insecure: true,
          exportGlobals: exportGlobal,
          timeoutScript:0
        },

**PFA:** summary json of the collection executed.

[New Compressed (zipped) Folder.zip](https://github.com/postmanlabs/newman/files/6569201/New.Compressed.zipped.Folder.zip)

vikyd:
this bug makes `sendRequest` unusable.

vbisbest:
@vikyd The behavior I am seeing is that I get duplicate main requests.  For instance, Newman shows

1. pre-request 1
2. main request 1
3. pre-request 2
4. main request 2

instead, the reporter is showing
1. main request 1
2. main request 1
3. main request 2
4. main request 2

Thank you.

vikyd:
> @vikyd The behavior I am seeing is that I get duplicate main requests. For instance, Newman shows
> 
> 1. pre-request 1
> 2. main request 1
> 3. pre-request 2
> 4. main request 2
> 
> instead, the reporter is showing
> 
> 1. main request 1
> 2. main request 1
> 3. main request 2
> 4. main request 2
> 
> Does this PR fix that? Thank you.

Yes. this PR fix that.

PR #2755, opened 2021-06-10
fix: pm.sendRequest summary incorrect duplicate item in run.executions

vikyd:
relative issues:

- https://github.com/postmanlabs/newman-reporter-html/issues/180
- https://github.com/postmanlabs/newman/issues/2752


byrdman:
Any chance this PR is going to be reviewed and merged?  Looks like at this point it's going to have to be caught up with the parent branch but this would be a useful fix.

PR #2762, opened 2021-06-12
Added Newman Request command, curl options, createCurl() and curl-to-postmanv2 plugin 

raghavbhatia336:
As a part of my Google Summer of Code project, I am creating a new command, `Newman Request` which: 
- allows the user to run a single request using Newman and show the request/response using a reporter of their choice. 
- replaces curl with newman request in curl command and runs and sends the request using Newman. 

To achieve these objectives, I plan to use [curl-to-postmanv2](https://github.com/postmanlabs/curl-to-postman) plugin to convert the command to postman collection with a single item. The created collection is then run using postman-runtime.

As the first two action items of the project, I have 
- added the newman request command using the commander library. 
- added the curl options supported by [curl-to-postman](https://github.com/postmanlabs/curl-to-postman) to the newman request command 

Supported curl options of curl-to-postmanv2:

- [x]  -A, --user-agent
- [x]  -d, --data
- [x]  --data-urlencode
- [x]  --data-binary
- [x]  --data-raw
- [ ]  --data-asci
- [x]  -F, --form (can be used multiple times)
- [x]  -G, --get
- [x]  -H, --header (can be used multiple times)
- [x]  -X, --request
- [x]  -I, --head
- [x]  -T, --file-upload
- [ ]  --url
- [ ]  --basic
- [ ]  -u, --user

I have also written the tests for some example commands: 

```
newman request -X GET http://www.google.com -r cli,json

newman request -X POST  https://postman-echo.com/post -H 'Content-Type: text/plain'  -r progress

newman request -X POST -F 'username=davidwalsh' -F 'password=something' https://postman-echo.com/post

newman request -X POST  https://postman-echo.com/post -d '{"hello":"world"}' -H 'Content-Type: text/json' -H 'Scheme: https' --user-agent 'mobile app' -r cli,progress

newman request -I https://postman-echo.com/post

newman request -G https://google.com
```

As the next two action items, I have:
- added the createCurl() function in ./bin/util.js that creates a stringified Curl command. 
- added curl-to-postmanv2 dependency to package.json. 
- written unit tests for the createCurl() function.


bin/newman.js
```diff
@@ -85,6 +85,42 @@ program
         });
     });
 
+// The `request` command allows you to run single request with the provided options.
+program
+    .command('request <url>')
+    .description('Run single request with provided options')
+    .usage('<url> [options]')
+    .option('-X, --request <method>',
```
raghavgarg1257, line 93: Just a caveat to be aware of while handling this option, `-X` is given preference over the native `--GET` option i.e. when both the options are provided, `--GET` will be ignored and the method in `-X` will be used.
```sh
# would resolve in PUT
$ curl --GET -X PUT https://google.com 

# would resolve in DELETE
$ curl --get -X DELETE https://google.com 
```

bin/newman.js
```diff
@@ -85,6 +85,42 @@ program
...
+        'Specify request method')
+    .option('-H, --header <header>',
+        'Specify request header', util.cast.memoize, [])
+    .option('-A, --user-agent <agent>',
+        'Specify request user agent')
+    .option('-d, --data <data>',
```
raghavgarg1257, line 99: Just a caveat to be aware of while handling this option, if no method is explicitly provided while using this option `-d`, the curl defaults to `POST` method, otherwise the provided method will be used.
```sh
# would resolve in PUT
$ curl -X PUT -d \'{"key":"value"}\' https://google.com 

# would resolve in POST
$ curl -d \'{"key":"value"}\' https://google.com 
```
**Note:** This applies to all the body/data related options.

raghavgarg1257 (changes requested):

bin/newman.js
raghavbhatia336: Noted.

bin/newman.js
raghavbhatia336: Noted.

bin/newman.js
```diff
@@ -85,6 +85,42 @@ program
...
+        'Sends the specified data to the server with type application/x-www-form-urlencoded or application/json')
+    .option('--data-raw <data>',
+        'Sends the specified data to the server in raw format')
+    .option('--data-urlencode <data>',
+        'Sends the specified data to the server with type application/x-www-form-urlencoded')
+    .option('--data-binary <data>',
+        'Sends the specified data as-is')
+    .option('-F, --form <name=content>',
+        'A single form-data field', util.cast.memoize, [])
+    .option('-G, --get',
```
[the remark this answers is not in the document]
raghavbhatia336: Noted, but curl is already supporting -g as another curl option, please check here: [Link](https://gist.github.com/eneko/dc2d8edd9a4b25c5b0725dd123f98b10#file-list-of-curl-options-txt-L62).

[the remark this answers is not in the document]
raghavgarg1257: Yup, this checks out. 👍

bin/util.js
```diff
@@ -1,4 +1,48 @@
-const _ = require('lodash');
+const _ = require('lodash'),
+    ALL_CURL_OPTIONS = {
```
raghavgarg1257, line 2: @raghavbhatia336 Let's move these constant configs to a separate file, for easier maintainability.

bin/util.js
```diff
@@ -101,6 +145,57 @@ module.exports = {
...
 
+    /**
+     * Extract curl options in the provided options.
+     * Create curl  command from the options.
+     *
+     * @param {Object} options - Commander.Command Instance
+     * @param {String} url - Requested url
+     * @returns {String} - Curl command
+     */
+    createCurl: (options, url) => {
```
raghavgarg1257, line 156: @raghavbhatia336 Let's identify if there could be a way to not create a tight coupling with curl options because it will affect our ability to support new options. Since we are just taking the options from the user and passing them onto the next module, we shouldn't really handle the options.
Can you please look into whether `program.args` solve our problem? This variable provides the user's command in an array form.

raghavgarg1257 (approved):

PR #2777, opened 2021-07-05
Added convert() function of curl-to-postmanv2 and added postman-runtime capability

raghavbhatia336:
In the next set of action items for , I have added:

- the _convert()_ function of `curl-to-postmanv2` which takes the stringified curl command and converts it into a Postman-Collection with a single item.

- the `postman-runtime` capability which is used to run the Postman-Collection.


bin/newman.js
```diff
@@ -112,10 +112,14 @@ program
         'Forces the request to be sent as HEAD, with the --data parameters appended to the query string')
     .option('-T, --upload-file <file>',
         'Forces the request to be sent as PUT with the specified local file to the server')
-    .action(() => {
-    /* eslint-disable no-console */
-        console.log('Newman Request is working!');
-    /* eslint-enable no-console */
+    .action((url, command) => {
+        let options = util.commanderToObject(command);
```
raghavgarg1257, line 116: Why are we using let for `options`? Are we expecting it to be changed further down the line?

test/unit/request-util.test.js
```diff
@@ -0,0 +1,33 @@
...
+            output: [{ type: 'request',
+                data: { method: 'GET',
+                    name: 'http://google.com',
+                    url: 'http://google.com',
+                    header: [],
+                    body: {},
+                    description: 'Generated from a curl request: \ncurl  --get http://google.com'
+                }
+            }
+            ] },
```
raghavgarg1257, line 16: Please use the correct indentation.

lib/request/index.js
```diff
@@ -0,0 +1,15 @@
+var runtime = require('postman-runtime'),
+    util = require('./util');
+
+const req = function (options, callback) {
+    var runner = new runtime.Runner();
+
+    const collection = util.convert(options.curl, callback);
```
raghavgarg1257, line 7: `util.convert` has an inconsistent return format.
- For successful processing, it is returning the data synchronously.
- For errors, it is calling the callbacks. Also which we didn't handled.

Please use a consistent return format. Since the `convert` method of the module is using a callback, we will have to use callback for all returning statements and handle it properly here.

lib/request/util.js
```diff
@@ -0,0 +1,20 @@
...
+        if (err) {
+            return callback(err);
+        }
+        collection.items.add({
+            name: result.output[0] && result.output[0].data.name,
+            request: result.output[0] && result.output[0].data
+        });
+    });
+
+    return collection;
```
raghavgarg1257, line 17: Even though the method `curl2postman.convert` is synchronous in execution nature but its implementation has been done in a way of the async method by the module owner and we will have to follow that. So, we can't return the `collection` from here.

We need to move this statement inside the callback of `curl2postman.convert` and return the collection via callback. For example `return callback(null, collection)`, `null` for no-error.

raghavgarg1257 (commented):
After the changes, we will address the test to increase the functional coverage to at least 55%.

PR #2785, opened 2021-07-12
Added multiple flag support in bin/util.js 

raghavbhatia336:
I have added the functionality in `createCurl()` function so that the following curl options have multiple-flag support: 
`-d, --data, --data-raw, --data-urlencoded`.


raghavgarg1257:
@raghavbhatia336 This was the temporary PR that is not needed anymore, right?

raghavbhatia336:
Yes, this can be closed now.

PR #2788, opened 2021-07-20
Added cli-output for single request 

raghavbhatia336:
In this action item, I have used Newman run capabilities to add cli output for a single request.


coditva (changes requested):
Looks good apart from few comments. Let's make it DRY soon and refactor out similar code into separate modules.

PR #2790, opened 2021-07-22
Fix Formatting of CLI Reporter Options Table in README

jenslauterbach:
There is a redundant newline in the READMEs CLI reporter options table. This newline causes the table formatting to break. The table is not displayed as a table at the moment.

This patch does remove the newline. If it is applied the table formatting will be ok again.


jenslauterbach:
Just to make it a bit clearer, what the current state is I attached a screenshot.

Screenshot 2021-07-22 at 17 08 30

PR #2796, opened 2021-07-24
Exposed an extra parameter for single-request to the cli reporter  

raghavbhatia336:
In this action item, I have added an extra parameter, `options.singleRequest` in collection options for the cli reporter. This has been done to make the cli reporter aware of single-request and provide the output for single-request differently. 
```
Example Command: 
newman request -X POST 'https://postman-echo.com/post' -x -d 'name=admin' -d 'shoesize=12'
```
Output: 
image

The output now shows additional information about the single request (DNS lookup time, Response Time, Data Received, First Byte Time) than what is shown by the curl command.


lib/reporters/cli/index.js
```diff
@@ -288,6 +299,31 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
...
+        const summaryTable = new Table({
+                style: { head: [] },
+                colWidths: [40]
+            }),
+            timings = run.timings,
+            transfers = run.transfers;
+
+        // add row to show total data received
+        // eslint-disable-next-line max-len
+        transfers && summaryTable.push([colors.green('Data received: %s (approx)'), util.filesize(transfers.responseTotal)]);
```
raghavgarg1257, line 312: @raghavbhatia336 Here we are not formatting the string but simply adding a row to the table. Hence there is no need for `%s`. An alternative to the above line could be
```js
summaryTable.push([colors.green('Data received:'), util.filesize(transfers.responseTotal) + ' (approx)']);
```

lib/reporters/cli/index.js
```diff
@@ -81,9 +81,20 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
         // for some reason, if there is no run summary, it is unexpected and hence don't validate this
         var run = this.summary.run;
 
+        // show the body of the response for single request
+        // eslint-disable-next-line max-len
+        if (options.singleRequest && run.executions && run.executions.length && run.executions[0].response && run.executions[0].response.text()) {
+            print(run.executions[0].response.text());
```
raghavgarg1257, line 87: @raghavbhatia336 
- Can we improve the output for the response text? Maybe we can provide a flag on basis of which we will prettify the output or not.
- Can we verify the output of different formats like XML, or blob? How is it looking in the CLI?
- What would happen for very long outputs?

lib/reporters/cli/index.js
```diff
@@ -81,9 +81,20 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
         // for some reason, if there is no run summary, it is unexpected and hence don't validate this
         var run = this.summary.run;
 
+        // show the body of the response for single request
+        // eslint-disable-next-line max-len
```
coditva, line 85: I'm a little concerned that we're adding so many `eslint-disable` comments. Can we break this line properly and avoid these comments? 🤔

lib/reporters/cli/index.js
```diff
@@ -81,13 +81,80 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
         // for some reason, if there is no run summary, it is unexpected and hence don't validate this
         var run = this.summary.run;
 
+        function formatXML (inputXML, indent) {
```
raghavgarg1257, line 84: Instead of writing our own implementation, can we please explore any existing package?
This module seems good, https://github.com/chrisbottin/xml-formatter

lib/reporters/cli/index.js
```diff
@@ -81,13 +81,80 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
...
+        }
+
         // show the body of the response for single request
         if (options.singleRequest &&
             run.executions &&
             run.executions.length &&
             run.executions[0].response &&
             run.executions[0].response.text()) {
-            print(run.executions[0].response.text());
+            if (run.executions[0].response.headers.all()
```
raghavgarg1257, line 141: You can get this information using the method `contentInfo` on the object of `PostmanResponse`, without any loop or extra code. 

For example, the statement like this
```js
run.executions[0].response.contentInfo()
```
would give output like this
```js
{
    mimeType: 'text',
    mimeFormat: 'json', // the key which we can use
    charset: 'utf8',
    fileExtension: 'json',
    fileName: 'response.json'
}
```
Ref: https://www.postmanlabs.com/postman-collection/Response.html#contentInfo

lib/reporters/cli/index.js
```diff
@@ -81,13 +81,80 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
...
+
+                print(prettyxml);
+            }
+
+            else if (run.executions[0].response.headers.all()
+                .find((element) => element.key === 'Content-Type') // eslint-disable
+                .value.includes('application/json')) {
+                const a = JSON.parse(run.executions[0].response.text());
+
+                print(JSON.stringify(a, undefined, '\t'));
```
raghavgarg1257, line 156: Let's use SPACES instead of TAB, so in order to have consistency across the board. We can keep it as 2 or 4 spaces, both are fine.

raghavgarg1257 (changes requested):

lib/reporters/cli/index.js
```diff
@@ -81,9 +81,39 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
...
+            run.executions &&
+            run.executions.length &&
+            run.executions[0].response &&
+            run.executions[0].response.text()) {
+            const xmlFormat = require('xml-formatter');
+            if (run.executions[0].response.contentInfo().mimeFormat === 'xml' || run.executions[0].response.contentInfo().mimeFormat === 'html' ) {
+                print(xmlFormat(run.executions[0].response.text()));
+            }
+
+            else if (run.executions[0].response.headers.all()
```
raghavgarg1257, line 100: Please replace all the usage of `response.header` with `response.contentInfo()`

lib/reporters/cli/index.js
```diff
@@ -81,9 +81,39 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
         // for some reason, if there is no run summary, it is unexpected and hence don't validate this
         var run = this.summary.run;
 
+        if (options.verbose) {
```
raghavgarg1257, line 84: Please move the verbose output after we show the default outputs.

lib/request/index.js
```diff
@@ -15,21 +30,76 @@ module.exports = function (options, callback) {
...
-            callback(err); // eslint-disable-line
+            return callback(err);
         }
 
-        runner.run(curlCollection, {}, function (err, runResult) {
-            console.log(err); // eslint-disable-line
-            console.log(runResult); // eslint-disable-line
+        runner.run(curlCollection, {
+            requester: {
+                verbose: true
```
raghavgarg1257, line 46: Please use verbose only when the user gives the flag, otherwise, we are just processing for more than what we need.

package.json
```diff
@@ -60,6 +60,7 @@
     "serialised-error": "1.1.3",
     "tough-cookie": "3.0.1",
     "word-wrap": "1.2.3",
+    "xml-formatter": "^2.4.0",
```
raghavgarg1257, line 63: Please remove the carat from the module version. We always use the fixed version for a package so that it automatically doesn't bump and surprise us in the production. We try to update only when required.

lib/reporters/cli/index.js
```diff
@@ -81,9 +81,39 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
...
+            print('\n\n');
+        }
+
+        // show the body of the response for single request
+        if (options.singleRequest &&
+            run.executions &&
+            run.executions.length &&
+            run.executions[0].response &&
+            run.executions[0].response.text()) {
+            const xmlFormat = require('xml-formatter');
```
raghavgarg1257, line 95: Let's move all the requires to the start of the file

lib/reporters/cli/index.js
```diff
@@ -288,6 +294,49 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
...
+
+        timingPhases && _.forEach({
+            prepare: 'Prepare:',
+            tcp: 'TCP:',
+            firstByte: 'FirstByte:',
+            download: 'Download:',
+            secureHandshake: 'SecureHandshake:'
+        }, (label, key) => {
+            timingPhases &&
+            summaryTable.push([colors.green(label), util.prettyms(timingPhases[`${key}`]) + 'ms']);
```
raghavgarg1257, line 334: `util.prettyms` is already appends the unit to the value. Currently it is printing like `56msms`

lib/reporters/cli/index.js
```diff
@@ -200,31 +230,7 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
         timings = _.last(_.get(o, 'history.execution.data'));
         timings = timings && timings.timings;
 
-        // print timing info of the request
```
raghavgarg1257, line 203: We cannot remove this part, because this is essential for `newman run`, we will have to put conditionals using the flag for `singleRequest`

raghavgarg1257: @raghavbhatia336 Just a thought, can we use this as default for showing timing phases instead of our own table? This looks neater to me.

raghavgarg1257 (changes requested):

raghavgarg1257:
@raghavbhatia336 Please resolve the conflicts which arose from the recent release of the newman.

raghavgarg1257 (approved):

PR #2806, opened 2021-08-03
This exposes the `runner` on newman and the runtime `run` object in start event

shamasis:
The change is tiny and the code-diff would be self explanatory.

This cannot be documented as of yet, since runtime itself does not have enough documentation to leverage extensibility. However, it is a start and in a few iterations, the documentation can be improved.

For now, with this change, powerful reporters such as GIF attached below can be built.
Aug-03-2021 23-38-06

### DIFF
--- a/lib/run/index.js
+++ b/lib/run/index.js
@@ -226,10 +226,9 @@ module.exports = function (options, callback) {
                  * @returns {*}
                  */
                 start (err, cursor) {
-                    emitter.emit('start', null, {
+                    emitter.emit('start', err, {
                         cursor: cursor,
-                        run: run,
-                        error: err
+                        run: run
                     });
                 },
 
--- a/lib/run/index.js
+++ b/lib/run/index.js
@@ -17,7 +17,6 @@ var _ = require('lodash'),
      * @type {Object}
      */
     runtimeEvents = {
-        start: [],
         beforeIteration: [],
         beforeItem: ['item'],
         beforePrerequest: ['events', 'item'],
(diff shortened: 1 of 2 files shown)


lib/run/index.js
```diff
@@ -215,6 +217,22 @@ module.exports = function (options, callback) {
...
+                 * @param {?Error} err - An Error instance / null object.
+                 * @param {Object} cursor - The run cursor instance.
+                 * @returns {*}
+                 */
+                start (err, cursor) {
+                    emitter.emit('start', null, {
+                        cursor: cursor,
+                        run: run,
+                        error: err
+                    });
```
coditva, line 233: Seems like a breaking change we can avoid. 😄 

```suggestion
                start (err, cursor) {
                    emitter.emit('start', err, {
                        cursor: cursor,
                        run: run
                    });
```

lib/run/index.js
```diff
@@ -156,6 +155,9 @@ module.exports = function (options, callback) {
         // to store the exported content from reporters
         emitter.exports = [];
 
+        // expose the runner object for reporter and programmatic use
+        emitter.runner = runner;
+
```
coditva, line 160: Sorry if it's a noob question, but why do we need to expose the `runner` object. AFAIK, we only need the `run` object (which we're exposing in `start` event) for programmatic access, right?

coditva:
cc @elit-altum
We can use the exposed `run` object here for commands in https://github.com/elit-altum/newman-dashboard. 🚀

lib/run/index.js
shamasis: Yes. I've exposed it in case right before start (at the construction point of the reporter), we need to configure the runner options or do some function wrapping shenanigans. 😛

lib/run/index.js
shamasis: True. Stupid me. 🙈

shamasis:
@coditva I have applied the recommended patches.

coditva (approved):
@shamasis Looks good!
@codenirvana can you take a final look and merge?! 🚀

codenirvana (approved):

PR #2808, opened 2021-08-05
Added additional CLI output

raghavbhatia336:


lib/reporters/cli/index.js
```diff
@@ -334,6 +349,17 @@ _.assignIn(PostmanCLIReporter, {
             summaryTable.push([colors.green(label), util.prettyms(timings[`${key}Average`])]);
         });
 
+        timingPhases && _.forEach({
+            prepare: 'Prepare:',
+            tcp: 'TCP:',
+            firstByte: 'FirstByte:',
```
raghavgarg1257, line 355: First byte data is duplicated.

lib/reporters/cli/index.js
```diff
@@ -334,6 +349,17 @@ _.assignIn(PostmanCLIReporter, {
             summaryTable.push([colors.green(label), util.prettyms(timings[`${key}Average`])]);
         });
 
+        timingPhases && _.forEach({
```
raghavgarg1257, line 352: Let's use `timingPhases` to show all the timings in the following order and remove the above `timings` variable usage.
```
    dns: Number,             // duration of DNS lookup
    tcp: Number,             // duration of TCP connection
    secureHandshake: Number, // duration of secure handshake
    firstByte: Number,       // duration of HTTP server response
    download: Number,        // duration of HTTP download
    total: Number            // duration entire HTTP round-trip
```
Ref: https://github.com/postmanlabs/postman-runtime/blob/develop/docs/history.md

raghavgarg1257 (changes requested):

lib/run/summary.js
```diff
@@ -319,6 +319,9 @@ _.assign(RunSummary, {
             timings = timings && timings.timings;
             timingPhases = timings && sdk.Response.timingPhases(timings);
 
+            summary.run.timingPhases = timingPhases;
```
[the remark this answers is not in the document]
raghavbhatia336: So the fields for run are given to be stats, timings, executions, failures, transfers and error. For request it would be timingPhases, executions, transfers, failures and error, right?

lib/reporters/cli/index.js
raghavbhatia336: Ok, will change.

raghavbhatia336: We still keep run.transfers right? It is being used for showing total data received.

raghavgarg1257: Yup, `run.transfers` will be kept.

raghavgarg1257 (approved):

PR #2811, opened 2021-08-10
Added response headers

raghavbhatia336:


lib/reporters/cli/index.js
```diff
@@ -90,14 +90,24 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
...
                 print(xmlFormat(run.executions[0].response.text()));
             }
 
             else if (run.executions[0].response.contentInfo().mimeFormat === 'json') {
-                const inputJSON = JSON.parse(run.executions[0].response.text());
+                const inputJSON = JSON.parse(run.executions[0].response.text()),
+                    responseHeaders = run.executions[0].response.headers.all();
 
                 print('\n');
+                responseHeaders.map((element) => {
```
raghavgarg1257, line 108: Why are we repeating the same code? Can't we DRY it?

PR #2818, opened 2021-08-17
Handling large data for single request

raghavbhatia336:
Added the capability to handle large sized data for single request.

### DIFF
--- a/lib/reporters/cli/index.js
+++ b/lib/reporters/cli/index.js
@@ -88,7 +88,8 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
             run.executions &&
             run.executions.length &&
             run.executions[0].response &&
-            run.executions[0].response.text()) {
+            run.executions[0].response.text() &&
+            run.transfers.responseTotal < 10485760) {
             print('\n');
             const responseHeaders = run.executions[0].response.headers.all();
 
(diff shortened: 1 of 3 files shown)


lib/reporters/cli/index.js
```diff
@@ -108,11 +109,20 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
...
+                LF +
+                'Please use the JSON reporter (-r JSON) to download the output in a separate file.');
+        }
+
         // show the summary table (provided option does not say it is not to be shown)
         if (!reporterOptions.noSummary) {
             if (options.singleRequest) {
-                print('\n');
-                print(LF + PostmanCLIReporter.parseSingleRequestStatistics(run) + LF);
+                if (run.transfers.responseTotal < 10485760) {
```
raghavgarg1257, line 122: We can show stats for large output sizes as well.

lib/reporters/cli/index.js
```diff
@@ -108,11 +109,20 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
             }
         }
 
+        if (run.transfers.responseTotal > 10485760) {
```
raghavgarg1257, line 112: Since the output size conditional only applies to output printing, so lets move this if conditional inside of the above if conditional.

lib/reporters/cli/index.js
```diff
@@ -88,7 +88,8 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
             run.executions &&
             run.executions.length &&
             run.executions[0].response &&
-            run.executions[0].response.text()) {
+            run.executions[0].response.text() &&
+            run.transfers.responseTotal < 10485760) {
```
raghavgarg1257, line 92: Points while using the numbers like these
- never hardcode them in the logic but always use config.
- never use the whole final number but how you reached at that number example `1024 * 1024 * 10`. This makes reading the code much easier.
- always add a comment on what that number signifies with correct unit, example // 10 MB

raghavgarg1257 (changes requested):

lib/reporters/cli/index.js
```diff
@@ -88,23 +88,32 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
...
+            if (run.transfers.responseTotal > (options.responseLimit)) {
+                print('\n');
+                print('Response Size is too large.' +
+                    LF +
+                    'Please use the JSON reporter (-r JSON) to download the output in a separate file.');
             }
 
-            else if (run.executions[0].response.contentInfo().mimeFormat === 'json') {
-                const inputJSON = JSON.parse(run.executions[0].response.text());
+            else if (run.transfers.responseTotal < (options.responseLimit)) {
```
raghavgarg1257, line 107: Why do we require this conditional, can't we simply use `else`?

raghavgarg1257 (approved):

PR #2819, opened 2021-08-17
New feature `newman request`

raghavgarg1257:
This was developed under the GSoC project by the student @raghavbhatia336.

### DIFF
--- a/bin/util.js
+++ b/bin/util.js
@@ -1,4 +1,5 @@
-const _ = require('lodash');
+const _ = require('lodash'),
+    { ALL_CURL_OPTIONS } = require('./constants.js');
 
 module.exports = {
 
@@ -99,6 +100,57 @@ module.exports = {
 
             return result;
         }, {});
+    },
+
+    /**
+     * Extract curl options in the provided options.
+     * Create curl  command from the options.
+     *
+     * @param {Object} options - Commander.Command Instance
+     * @param {String} url - Requested url
+     * @returns {String} - Curl command
+     */
+    createCurl: (options, url) => {
+        // Get all curl option names
+        const allCurlOptions = Object.keys(ALL_CURL_OPTIONS),
+            curlOptions = _.reduce(options, (result, value, key) => {
+                // Exclude non curl options
+                const validProp = _.includes(allCurlOptions, key);
+
+                validProp && (result[key] = value);
+
+                return result;
+            }, {}),
+
+            curlOptionToString = (option, value) => {
+                if (option.format) {
+                    return `${option.long} '${value}'`;
+                }
+
+                return `${option.long}`;
+            },
+
+            headers = _.reduce(curlOptions.header, (result, value) => {
+                return result + `-H '${value}'`;
+            }, ''),
+
+            forms = _.reduce(curlOptions.form, (result, value) => {
+                return result + `-F '${value}'`;
+            }, ''),
+
+            command = Object.keys(curlOptions).reduce((memo, optionName) => {
+                // forms and headers are collected and added separately
+                if (optionName === 'header' || optionName === 'form') {
+                    return memo;
+                }
+
+                const option = ALL_CURL_OPTIONS[optionName],
+                    value = curlOptions[optionName];
+
+                return memo + ` ${curlOptionToString(option, value)}`;
+            }, 'curl');
+
+        return `${command} ${url} ${headers} ${forms}`;
     },
 
     /**
--- a/bin/util.js
+++ b/bin/util.js
@@ -111,10 +111,12 @@ module.exports = {
      * @returns {String} - Curl command
      */
     createCurl: (options, url) => {
-        // Get all curl option names
-        const allCurlOptions = Object.keys(ALL_CURL_OPTIONS),
+        const
+            // Get all curl option names
+            allCurlOptions = Object.keys(ALL_CURL_OPTIONS),
+
+            // Exclude non curl options
             curlOptions = _.reduce(options, (result, value, key) => {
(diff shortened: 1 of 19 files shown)


bin/util.js
```diff
@@ -101,6 +102,58 @@ module.exports = {
...
+    createCurl: (options, url) => {
+        const
+            // Get all curl option names
+            allCurlOptions = Object.keys(ALL_CURL_OPTIONS),
+
+            // Exclude non curl options
+            curlOptions = _.reduce(options, (result, value, key) => {
+                const validProp = _.includes(allCurlOptions, key);
+
+                validProp && (result[key] = value);
```
coditva, line 122: This check can move to the `map` on line 150 itself, I think.

bin/util.js
```diff
@@ -101,6 +102,58 @@ module.exports = {
...
+
+                    return optionsValue;
+                }
+
+                if (curlOption.format && userOptionValue.length > 0) {
+                    return `${curlOption.long} '${userOptionValue}'`;
+                }
+
+                if (!curlOption.format) {
+                    return `${curlOption.long}`;
```
coditva, line 144: Why wrap in template string? 🤔

lib/reporters/cli/index.js
```diff
@@ -81,15 +83,55 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
         // for some reason, if there is no run summary, it is unexpected and hence don't validate this
         var run = this.summary.run;
 
+        // show the body of the response for single request
+        if (options.singleRequest &&
```
coditva, line 87: We need a cleaner way to add branches for `singleRequest` and `run`. Probably separate it out into two different functions and call either of them based on the flag.

Currently we're adding these checks in an ad-hoc manner. Plus, it becomes difficult to add changes to `run` output without affecting `request` output.

lib/request/index.js
```diff
@@ -0,0 +1,105 @@
+const _ = require('lodash'),
```
coditva, line 1: Need to remove duplication with `lib/run/index.js` file.

We should break this and `run/index.js` file down into smaller functions with clear interfaces and update these files to use the functions.

lib/request/index.js
```diff
@@ -0,0 +1,105 @@
...
+module.exports = function (options, callback) {
+    // validate all options. it is to be noted that `options` parameter is option and is polymorphic
+    (!callback && _.isFunction(options)) && (
+        (callback = options),
+        (options = {})
+    );
+    !_.isFunction(callback) && (callback = _.noop);
+
+    // ensure that the curl command is present before starting a run
+    if (!_.isString(options.curl)) {
```
coditva, line 29: I'm not sure this is what we want... Newman is not just a CLI tool but also a library. If the user has to pass the curl command as a string, then it sets a different expectation than the `run` function on the library where the user can pass options object.

I think we should move the conversion of options to string here instead of the `bin/newman.js`. Benefits:
1. Consistent interface like `run`.
2. Moving the business logic to library instead of in the binary which should only be responsible to add a executable layer on top of the library. The "meat" of the code should lie in the library.

lib/request/index.js
```diff
@@ -0,0 +1,105 @@
...
+
+                // we now trigger actual done event which we had overridden
+                emitter.emit('done', null, emitter.summary);
+                callback(null, emitter.summary);
+            };
+
+            // initialise all the reporters
+            !emitter.reporters && (emitter.reporters = {});
+            if (!emitter.reporters.cli) {
+                emitter.reporters.cli = new CliReporter(emitter, {}, options);
```
coditva, line 96: Need to be changed to allow all reporters to work here.

lib/request/index.js
coditva: Another point to note is that in future we would not need to convert it to string anymore. If we keep this as it is, we will either have to break the API to now send options rather than the string, or support both APIs where we accept string as well as options object. 😬

lib/run/summary.js
```diff
@@ -319,6 +373,9 @@ _.assign(RunSummary, {
             timings = timings && timings.timings;
             timingPhases = timings && sdk.Response.timingPhases(timings);
 
+            summary.run.timingPhases = timingPhases;
+            summary.run.verboseSession = _.findLast(_.get(o, 'history.execution.sessions'));
```
coditva, line 377: Do you mean `_.last`?
```suggestion
            summary.run.verboseSession = _.last(_.get(o, 'history.execution.sessions'));
```

coditva (changes requested):
Have not reviewed tests yet. Left some comments for the changes.

lib/request/index.js
raghavbhatia336: This change has been incorporated. Thanks!

lib/reporters/cli/index.js
raghavbhatia336: This change has been incorporated. Thanks!

lib/request/index.js
raghavbhatia336: Resolved.

raghavgarg1257:
Hello @shamasis @codenirvana, 👋 
Kindly review the PR for the new feature Newman Request and please share your thoughts on wherever we can improve this feature or code for better maintainability.
The student has also created a PR for documentation #2830. This is not yet merged in the feature branch for it still needs some work, and we are open to suggestions for any improvement in the doc.
Thanks. 🙏

shamasis:
Raghav ... we might have to prioritise some exigent items the team is working on at the moment. Someone will get back on this ASAP.

PR #2822, opened 2021-08-20
Separated code of cli-reporter for request and run

raghavbhatia336:


raghavgarg1257:
@raghavbhatia336 Can we rename the files to have the entry point to be `lib/reporters/cli/index.js` and have the command name in the name of its files like `cli-request.js` and `cli-run.js`, for this will help in the consistency, so that in future if someone has to add a separate file for a new command they can simply create the file `cli-{command}.js` and use it in `cli/index.js`. Also common node js practice says that `index`, should always be the entry point for a directory.

raghavgarg1257:
Merged this branch without all the checks because integration tests are only covering the command `newman run` and with every new file added for the new command `newman request`, the coverage is decreasing.  Right now the coverage for the integration test looks like this:
```
45 integrations ok!

ERROR: Coverage for functions (34.67%) does not meet global threshold (35%)
ERROR: Coverage for branches (19.63%) does not meet global threshold (20%)
```

There are a couple of approaches we can take and create a new PR for handling integration tests.
- Write integration test for `newman request`, which might not be fruitful for these test mostly will be covering the module `curl-to-postmanv2`'s functionality and not our code.
- Find and DRY the duplicate "functions" between `newman run` and `newman request`, so that total functions get reduced and coverage percent will increase as the result.
- Decrease the threshold

PR #2849, opened 2021-09-27
Upload newman runs to Postman

UmedJadhav:
<TBA>

### DIFF
--- a/bin/newman.js
+++ b/bin/newman.js
@@ -1,4 +1,6 @@
 #!/usr/bin/env node
+/* eslint-disable function-paren-newline */
+/* eslint-disable quotes */
 
 require('../lib/node-version-check'); // @note that this should not respect CLI --silent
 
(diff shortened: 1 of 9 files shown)


lib/run/index.js
```diff
@@ -340,10 +349,13 @@ module.exports = function (options, callback) {
                         });
                     });
 
-                    asyncEach(emitter.exports, exportFile, function (err) {
+                    asyncEach(emitter.exports, exportFile, async function (err) {
                         // we now trigger actual done event which we had overridden
                         emitter.emit('done', err, emitter.summary);
-                        callback(err, emitter.summary);
+
+                        var resultUploaded = await new uploadRunToPostman(uploadConfig, emitter.summary).start();
```
coditva, line 356: This could be just a method call instead if we only ever need one method from the instance of the class. Why create a class for this at all?

lib/run/upload-run.js
```diff
@@ -0,0 +1,133 @@
+const util = require('../util');
+
+const _ = require('lodash'),
+print = require('../print'),
+retry = require('async-retry'),
+request = require('postman-request'),
+ERROR_GO_TO_LINK = '<TBA>',
+UPLOAD_RUN_API_URL = '<TBA>'
+
+class uploadRunToPostman{
```
coditva, line 10: We don't need to create a class for this. Class is needed only if you want to abstract some data and use the instance in multiple places. In our case, we only need a single method.

Sidenote: The name of the class always starts with a capital letter.

lib/run/upload-run.js
```diff
@@ -0,0 +1,133 @@
+const util = require('../util');
+
+const _ = require('lodash'),
```
coditva, line 3: Multiple `const` declarations. Can be combined into one.

lib/run/upload-run.js
```diff
@@ -0,0 +1,133 @@
...
+     * @returns {function} Returns an async function which can be used by async-retry library to have retries
+     */
+     _upload = (uploadOptions) => {
+        return async(bail) => await new Promise((resolve, reject) => request.post(uploadOptions , (error, response, body) => {
+                    if(error){
+                        if(error.code === 'ECONNREFUSED') return reject(error) // Retry only if the ERROR is ERRCONNECT
+                        return bail(error); // For other errors , dont retry
+                    }
+
+                    if(200 <= response.statusCode && response.statusCode <= 299){
```
coditva, line 39: This seems weird. We should know the discreet status codes that are returned by the API and check on those instead of just checking the range.

lib/run/upload-run.js
```diff
@@ -0,0 +1,133 @@
...
+
+    /**
+     * @private
+     * Internal upload call
+     *
+     * @param {Object} uploadOptions
+     * @returns {function} Returns an async function which can be used by async-retry library to have retries
+     */
+     _upload = (uploadOptions) => {
+        return async(bail) => await new Promise((resolve, reject) => request.post(uploadOptions , (error, response, body) => {
```
coditva, line 33: If you're already returning a Promise, you don't need to `await`.

```suggestion
        return (bail) => new Promise((resolve, reject) => request.post(uploadOptions , (error, response, body) => {
```

lib/run/upload-run.js
```diff
@@ -0,0 +1,133 @@
...
+    buildRunObject = ( runSummary, collectionUID , environmentUID ) => {}
+
+    /**
+     * @private
+     * Internal upload call
+     *
+     * @param {Object} uploadOptions
+     * @returns {function} Returns an async function which can be used by async-retry library to have retries
+     */
+     _upload = (uploadOptions) => {
```
coditva, line 32: Why does it need to return a function? Couldn't this function itself be used in `async/retry`?

lib/run/upload-run.js
```diff
@@ -0,0 +1,133 @@
...
+                    if(error){
+                        if(error.code === 'ECONNREFUSED') return reject(error) // Retry only if the ERROR is ERRCONNECT
+                        return bail(error); // For other errors , dont retry
+                    }
+
+                    if(200 <= response.statusCode && response.statusCode <= 299){
+                        return resolve(body);
+                    }
+
+                    if( 400 <= response.statusCode && response.statusCode<= 499){
```
coditva, line 43: Similar as above:
> We should know the discreet status codes that are returned by the API and check on those instead of just checking the range.

lib/run/upload-run.js
```diff
@@ -0,0 +1,133 @@
...
+
+                        if(response.statusCode === 404){
```
coditva, line 45: Inconsistent spacing.
```suggestion
                        if (response.statusCode === 404) {
```

lib/run/upload-run.js
```diff
@@ -0,0 +1,133 @@
...
+                            return bail(new Error('Couldn\'t find the postman server route'))
+                        }
+
+                        return bail(new Error(body.message)); // Avoid retry if there is client side error ( API key error / Workspace ID / permission error)
+
+                    }
+
+                    if( 500 <= response.statusCode && response.statusCode <= 599){ // Perform Retry if Server side Error
```
coditva, line 53: Similar as above:

> We should know the discreet status codes that are returned by the API and check on those instead of just checking the range.

lib/run/upload-run.js
```diff
@@ -0,0 +1,133 @@
...
+    /**
+     * @private
+     * Internal upload function which handles the retry
+     *
+     * @param {*} uploadOptions
+     * @param {*} retryOptions
+     * @returns {Promise}
+     */
+    _uploadWithRetry = (upload, retryOptions) => {
+        return retry( upload,{
```
coditva, line 71: Weird spacing.
```suggestion
        return retry(upload, {
```

lib/run/upload-run.js
```diff
@@ -0,0 +1,133 @@
...
+
+    /**
+     * @public
+     * Starts the run upload process
+     *
+     * @returns {Boolean}  Indicate if run upload was successful
+     */
+    start = async () => {
+        if(!this.uploadConfig.publishWorkspace){
+            return true;
```
coditva, line 89: Shouldn't this check be made before calling the `start` method itself? 🤔

Also, since this is an async function. It would be better to reject the returned promise by `throw`-ing here (and everywhere you want to return `false`) with appropriate error message.

lib/run/upload-run.js
```diff
@@ -0,0 +1,133 @@
...
+        };
+
+        try{
+            const response = await this._uploadWithRetry(this._upload(uploadOptions), retryOptions)
+
+            print.lf(`Uploaded the newman run to postman.
+                    Visit ${response.message} to view the results in postman web`);
+            return true;
+
+        } catch(error) {
```
coditva, line 124: Let the caller catch the error and show appropriate message.

lib/run/upload-run.js
```diff
@@ -0,0 +1,133 @@
...
+            return false;
+        }
+
+        print.lf('Uploading newman run to Postman');
+
+        const run = this.buildRunObject(runSummary, uploadConfig.publishWorkspaceSkipResponse);
+
+        const uploadOptions = {
+            url: UPLOAD_RUN_API_URL,
+            body: JSON.stringify(run),
```
coditva, line 103: `JSON.stringify` will throw in some cases. Verify that those cases won't ever occur or catch the error to handle those cases.

lib/run/upload-run.js
```diff
@@ -0,0 +1,133 @@
...
+            headers: {
+                'Content-Type': 'application/json',
+                'X-API-Header': uploadConfig.postmanApiKey
```
coditva, line 106: Is this the correct key for key header? Shouldn't it be `X-Api-Key`? 🤔

coditva (changes requested):

lib/run/upload-run.js
UmedJadhav: Added a try catch block

bin/newman.js
```diff
@@ -139,7 +143,7 @@ function run (argv, callback) {
 
 // This hack has been added from https://github.com/nodejs/node/issues/6456#issue-151760275
 // @todo: remove when https://github.com/nodejs/node/issues/6456 has been fixed
-(Number(process.version[1]) >= 6) && [process.stdout, process.stderr].forEach((s) => {
+Number(process.version[1]) >= 6 && [process.stdout, process.stderr].forEach((s) => {
```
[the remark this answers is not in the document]
UmedJadhav: Prettier did it , reversing it

bin/newman.js
```diff
@@ -61,9 +61,12 @@ program
     .option('--cookie-jar <path>', 'Specify the path to a custom cookie jar (serialized tough-cookie JSON) ')
     .option('--export-cookie-jar <path>', 'Exports the cookie jar to a file after completing the run')
     .option('--verbose', 'Show detailed information of collection run and each request sent')
+    .option('-p, --publish-workspace <workspace-id>', 'Publishes to given workspace')
+    .option('--publish-workspace-skip-response', 'Skip responses (headers, body, etc) while uploading newman run to Postman')
+    .option('--publish-retry', 'Number of times newman can try to publish before safely erroring out.')
```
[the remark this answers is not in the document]
UmedJadhav: Makes sense

bin/newman.js
```diff
@@ -61,9 +61,12 @@ program
     .option('--cookie-jar <path>', 'Specify the path to a custom cookie jar (serialized tough-cookie JSON) ')
     .option('--export-cookie-jar <path>', 'Exports the cookie jar to a file after completing the run')
     .option('--verbose', 'Show detailed information of collection run and each request sent')
+    .option('-p, --publish-workspace <workspace-id>', 'Publishes to given workspace')
```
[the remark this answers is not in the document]
giridharvc7: and `--publish-skip-response`

bin/newman.js
```diff
@@ -74,14 +77,15 @@ program
...
+        newman.run(options, function (err, summary, newmanStatus) {
             const runError = err || summary.run.error || summary.run.failures.length;
 
             if (err) {
                 console.error(`error: ${err.message || err}\n`);
                 err.friendly && console.error(`  ${err.friendly}\n`);
             }
-            runError && !_.get(options, 'suppressExitCode') && process.exit(1);
+
+            runError && !newmanStatus.resultUploaded && !_.get(options, 'suppressExitCode') && process.exit(1);
```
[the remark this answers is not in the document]
coditva: This should be an `||`.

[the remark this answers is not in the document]
coditva:
```suggestion
            (runError || !newmanStatus.resultUploaded) && !_.get(options, 'suppressExitCode') && process.exit(1);
```

lib/run/index.js
coditva:
```suggestion
                        new uploadRunToPostman(uploadConfig, emitter.summary).start().then((resultUpdated) => 
```

lib/run/upload-run.js
coditva: Hint: [.bind()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind)

lib/run/upload-run.js
UmedJadhav: Fixed

lib/run/upload-run.js
UmedJadhav: Moved the check to `run/index.js`

lib/run/upload-run.js
UmedJadhav: Reworked the flow

bin/newman.js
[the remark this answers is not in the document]
UmedJadhav: Fixed this

bin/newman.js
```diff
@@ -61,9 +61,12 @@ program
     .option('--cookie-jar <path>', 'Specify the path to a custom cookie jar (serialized tough-cookie JSON) ')
     .option('--export-cookie-jar <path>', 'Exports the cookie jar to a file after completing the run')
     .option('--verbose', 'Show detailed information of collection run and each request sent')
+    .option('-p, --publish-workspace <workspace-id>', 'Publishes to given workspace')
+    .option('--publish-workspace-skip-response', 'Skip responses (headers, body, etc) while uploading newman run to Postman')
```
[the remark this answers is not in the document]
UmedJadhav: Changed all the CLI parameter names

lib/run/index.js
```diff
@@ -302,7 +304,10 @@ module.exports = function (options, callback) {
                     // in case runtime faced an error during run, we do not process any other event and emit `done`.
                     // we do it this way since, an error in `done` callback would have anyway skipped any intermediate
                     // events or callbacks
-                    if (err) {
+
+                    // TODO: Add analytics call here
+
+                    if (err) { // TODO -> understand when this is raised
```
coditva, line 310: This error is raised from the `postman-runtime` library if the collection run did not finish successfully.

lib/run/upload-run.js
```diff
@@ -0,0 +1,257 @@
...
+/**
+ * @private
+ * Internal upload call
+ *
+ * @param {Object} uploadOptions
+ * @returns {function} Returns an async function which can be used by async-retry library to have retries
+ */
+_upload = (uploadOptions) => {
+    return async (bail) => request.post(uploadOptions , (error, response, body) => {
+                if(error){
```
coditva, line 21: Indentation looks weird here.

lib/run/upload-run.js
```diff
@@ -0,0 +1,257 @@
...
+                    if(error.code === 'ECONNREFUSED') { // Retry only if the ERROR is ERRCONNECT
+                        throw new Error(error.message);
+                    }
+                    return bail(error); // For other errors , dont retry
+                }
+
+                // Handle exact status codes
```
coditva, line 28: We don't seem to be handling the exact status codes. 😅

lib/run/upload-run.js
```diff
@@ -0,0 +1,257 @@
...
+    NEWMAN_STRING = 'newman';
+
+/**
+ * @private
+ * Internal upload call
+ *
+ * @param {Object} uploadOptions
+ * @returns {function} Returns an async function which can be used by async-retry library to have retries
+ */
+_upload = (uploadOptions) => {
```
coditva, line 19: If this is going to return a function, it should be called something like `_getUploadRetryable`.
Or, don't return a function! Just bind the params where you're using it.

coditva (changes requested):

PR #2852, opened 2021-10-04
by using base alpine (instead of node alpine) and adding npm and newm…

aaronhmiller:
…an. image size goes from 137MB to 109MB, shrinking it by 20%.


aaronhmiller:
Looks like this won't be merged. Closing to focus on active issues.

PR #2864, opened 2021-10-26
Fixed failed iterations count is always zero - even if an iteration had assertion errors #1600

### LINKED ISSUE #1600: Failed iterations count is always zero - even if an iteration had assertion errors
Hello there,

Version and environment information:

 1. Newman Version: 3.9.4
 2. OS details: Mac and CentOS
 3. Are you using Newman as a library, or via the CLI? Library
 4. Did you encounter this recently, or has this bug always been there: I think it has always been there
 
**5. Expected behaviour:**

When assertions within an iteration fail, I expect the iteration fail count to increase. Instead, it remains at 0.
This way, I cannot count successful iterations, only "run" iterations. I use the following variables:
* summary.run.stats.iterations.total
* summary.run.stats.iterations.failed

Additionally, the CLI report of test iterations displays this behavior (see attached screen shot): at least one iteration had failed assertions, but the failed iteration count is 0.

image001

Is this a bug or is my understanding faulty?

Best regards
Norman

pulkit-30:
Fix #1600 

### About Issue

When assertions within an iteration fail, the iteration fail count doesn't increase, it remains at 0.

### Fixed Solution 

Now, `summary.run.stats.iterations.failed` increase the value to 1  for the Assertion error.

### Code change ScreenShot

Screenshot 2021-10-26 at 4 36 24 PM

### Cli Report changes 

additionally, the CLI report of test iterations displays this behavior (see attached screenshot): Now the failed iteration count value is increased to 1 rather than showing 0.

Screenshot 2021-10-26 at 4 21 06 PM

What side effects does this change have?
This issue doesn't have any side effects. It passes all the test cases.

### DIFF
--- a/lib/run/summary.js
+++ b/lib/run/summary.js
@@ -390,6 +390,9 @@ _.assign(RunSummary, {
 
                 // if this is a plain error, convert it to serialised error
                 if (err.stack && !err.stacktrace) {
+                    if (err.name === 'AssertionError') {
+                        summary.run.stats.iterations.failed += 1;
+                    }
                     err = new SerialiseError(err, true);
                 }
 


shamasis (commented):
At a quick glance, I wasn't able to place why this is needed within an if-block whose generic job is to convert non type casted error.

My hunch is that change maybe needed to be done at a different place.

PR #2883, opened 2021-12-08
Added flag to set verbose output size limit for newman run

### LINKED ISSUE #2845: Verbose truncates output down to 2.05kB
1. Newman Version (can be found via `newman -v`): 5.3.0
 2. OS details (type, version, and architecture): MacOS 10.15.7
 3. Are you using Newman as a library, or via the CLI? CLI
 3. Did you encounter this recently, or has this bug always been there: Always been there (but I am new to this)
 4. Expected behaviour: full output when --verbose is used.
```
newman run NSX-T\ Create\ Edge\ Cluster.postman_collection.json  --verbose --reporters cli --insecure --environment NSX-T\ BED-LAB.postman_environment.json --env-var "nsx_manager_username=$NSXT_USERNAME" --env-var "nsx_manager_password=$NSXT_PASSWORD" --env-var "vcenter_username=$VCENTER_USERNAME" --env-var 'vcenter_password=$VCENTER_PASSWORD' --env-var "admin_user=$NSX_EDGE_VM_ADMIN_USERNAME" --env-var "admin_pass=$NSX_EDGE_VM_ADMIN_PASSWORD" --env-var "root_pass=$NSX_EDGE_VM_ROOT_PASSWORD" --env-var "snmpcommunity=$NSX_EDGE_VM_SNMP_COMMUNITY" --env-var "cluster=$NSX_EDGE_CLUSTER_SPEC"
```
A bash script runs the newman command after collecting the variables like credentials from the operator interactively and loading the remaining ones from files on disk
Screen Shot 2021-09-22 at 9 48 16 AM

Sid200026:
# Flag for Verbose Output Size Limit

Previously, the output limit when the verbose flag was set was 2KB or 2048 bytes which were hardcoded in the codebase. The main argument that can be made for hardcoding the limit to 2KB is large data streams can crash the terminal. However, this limit is unfair in cases when the output size is just beyond the 2KB size limit. Moreover in cases when you want to pipe the output of newman run to a file, setting a size limit doesn't make sense and moreover, the terminal won't crash.

This PR adds two new flags under --reporter-cli-*

1. --reporter-cli-verbose-output-limit <size in any unit> : Specify the truncation size of body in verbose CLI output for requests and responses
2. --reporter-cli-no-verbose-output-limit : This removes any truncation size limit of body in verbose CLI output for requests and responses
3. 
### Syntax

```
newman run <collection> --verbose --reporter-cli-verbose-output-limit  <size in any unit>
```
```
newman run <collection> --verbose --reporter-cli-no-verbose-output-limit
```

### Changes

- [x] Modified cli core logic
- [x] Added tests
- [x] Updated Documentation

Closes #2845


README.md
```diff
@@ -288,6 +291,7 @@ return of the `newman.run` function is a run instance, which emits run events th
 | options.timeout           | Specify the time (in milliseconds) to wait for the entire collection run to complete execution.<br /><br />_Optional_<br />Type: `number`, Default value: `Infinity` |
 | options.timeoutRequest    | Specify the time (in milliseconds) to wait for requests to return a response.<br /><br />_Optional_<br />Type: `number`, Default value: `Infinity` |
 | options.timeoutScript     | Specify the time (in milliseconds) to wait for scripts to return a response.<br /><br />_Optional_<br />Type: `number`, Default value: `Infinity` |
+| options.outputSize     | Specify the output size limit ( in KB ) when verbose flag is set.<br /><br />For example, `outputSize = 2` sets the verbose output limit to 2KB or 2048 bytes.<br /><br />To remove any verbose output limit, pass `infinite` to `--output-size`. This would set `outputSize` to `Number. POSITIVE_INFINITY`<br /><br />_Optional_<br />Type: `number`, Default value: `2` |
```
shamasis, line 294: Is KB the right unit? Any other option around size that we can draw pattern from?

bin/newman.js
```diff
@@ -50,6 +50,7 @@ program
     .option('--timeout [n]', 'Specify a timeout for collection run (milliseconds)', util.cast.integer, 0)
     .option('--timeout-request [n]', 'Specify a timeout for requests (milliseconds)', util.cast.integer, 0)
     .option('--timeout-script [n]', 'Specify a timeout for scripts (milliseconds)', util.cast.integer, 0)
+    .option('--output-size [n]', 'Specify the output size limit for CLI ( in  KB )', util.cast.integerOrInfinity, 2)
```
shamasis, line 53: English here is ambiguous "output size of what? Entire CLI, body always, body in verbose"?!

bin/util.js
```diff
@@ -20,6 +20,22 @@ module.exports = {
...
 
+        /**
+         * Helper to coerce number like strings into integers.
+         * If supplied argument is infinite, return positive infinity
+         * Perform safety checks, and return the result.
+         *
+         * @param {String} arg - The stringified number argument or infinite.
+         * @returns {Number} - The supplied argument, casted to an integer or positive infinity.
+         */
+        integerOrInfinity: (arg) => {
```
shamasis, line 31: How does Newman handle other properties that can have infinity? Any pattern to draw from there?

test/cli/verbose.test.js
```diff
@@ -33,4 +37,37 @@ describe('newman run --verbose', function () {
             done();
         });
     });
+
+    it('should limit to 2KB with --verbose option only without setting output-size', function (done) {
```
shamasis, line 41: Very neat and creative test. Like it ❤️😊. 

Another improvement could be if the body was autogenerated series of number marking increasing size. That way, we can both assert points and accuracy.

shamasis (changes requested):
Also, this codecov seems to be failing. We did cross the threshold in the PR, no clue why it's still failing.

README.md
```diff
@@ -174,6 +174,9 @@ For more details on [Reporters](#reporters) and writing your own [External Repor
 - `--timeout-script <ms>`<br />
   Specify the time (in milliseconds) to wait for scripts to complete execution.
 
+- `--output-size <KB>`<br />
```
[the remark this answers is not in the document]
Sid200026: Having gone through the reporter-cli params, I think bracketing it under reporter-cli-* would be more appropriate
How about replacing the current output-size flag with two reporter-cli flags
- `reporter-cli-verbose-output-limit <Size in any memory unit>`
- `reporter-cli-no-verbose-output limit` : Sets the limit to Infinite

It's rare to see Command Line tools allow users to pass arguments as infinite. Plus spelling errors might creep in.Mostly, they achieve by simply setting some flag.

README.md
Sid200026: Sure, will implement this. In case of invalid inputs, should we proceed with default 2KB or throw an error?

bin/newman.js
Sid200026: Technically, this PR limits the verbose body output only. I'll make the necessary changes to remove this ambiguity.

README.md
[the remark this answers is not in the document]
Sid200026: Hey @shamasis, any thoughts regarding this?

Sid200026:
@shamasis can you provide me with some additional feedback so that I can continue working on this?

Sid200026:
Hey @shamasis, @codenirvana @coditva, I have updated this PR with the necessary changes.

Currently there are 2 flags,
1. ```newman run <collection> --verbose --reporter-cli-verbose-output-limit  <size in any unit>```
2. ```newman run <collection> --verbose --reporter-cli-no-verbose-output-limit```

The --reporter-cli-verbose-output-limit uses [bytes](https://www.npmjs.com/package/bytes) to convert string input to bytes. In case of any invalid input, a warning is shown and a fallback size of 2KB is used.

lib/reporters/cli/index.js
```diff
@@ -26,7 +27,7 @@ var _ = require('lodash'),
         process: 'process',
         total: 'total'
     },
-    BODY_CLIP_SIZE = 2048,
+    DEFAULT_VERBOSE_OUTPUT_TRUNCATION_LIMIT = '2KB',
```
shamasis, line 30: Let's keep the constant name to body clip size. This way there is no ambiguity as to what output it is clipping.

lib/reporters/cli/index.js
```diff
@@ -70,6 +71,8 @@ extractSNR = function (executions) {
  * @param {Boolean=} reporterOptions.noFailures - Boolean flag to turn off failure reporting altogether, if set to true.
  * @param {Boolean=} reporterOptions.noConsole - Boolean flag to turn off console logging, if set to true.
  * @param {Boolean=} reporterOptions.noBanner - Boolean flag to turn off newman banner, if set to true.
+ * @param {String=} reporterOptions.verboseOutputLimit - Specify the truncation size of body in verbose CLI output.
```
shamasis, line 74: We do not need to introduce two parameters for this. In the grand scheme of things this is a small feature and one parameter should do. Here's what we can do:

1. We ALWAYS clip. No matter what.
2. Providing `Infinity` as an input option can then automatically prevent clipping with least complicated codes. (And we mention this behaviour in docs as a short byline.)
3. We take only one option `--reporter-cli-truncate-body-output` (people will 99% chance refer documentation to get to this, hence we name accordingly.)

Check for isNaN much earlier and prevent NaN from flowing too much down the code flow.

lib/reporters/cli/index.js
```diff
@@ -233,9 +236,15 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
                 `${mime.mimeType}`,
                 `${mime.mimeFormat}`,
                 `${mime.charset}`
-            ].join(` ${colors.gray(symbols.star)} `);
-
+            ].join(` ${colors.gray(symbols.star)} `),
+            // eslint-disable-next-line max-len
+            bodyClipSize = reporterOptions.verboseOutputLimit ? bytes(reporterOptions.verboseOutputLimit) : bytes(DEFAULT_VERBOSE_OUTPUT_TRUNCATION_LIMIT);
```
shamasis, line 241: Set the default value here. No need to compute. Instead at the below `if` block, set it if `!isNaN` and warn user in else of that block

lib/reporters/cli/index.js
```diff
@@ -249,13 +258,12 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
...
         ].join(` ${colors.gray(symbols.star)} `));
-
         // print request body
         if (reqTextLen) {
-            // truncate very large request (is 2048 large enough?)
-            if (reqTextLen > BODY_CLIP_SIZE) {
-                reqText = reqText.substr(0, BODY_CLIP_SIZE) +
-                    colors.brightWhite(`\n(showing ${util.filesize(BODY_CLIP_SIZE)}/${util.filesize(reqTextLen)})`);
+            // truncate very large request (if noVerboseOutputLimit flag is set, remove any limitations)
+            if (!reporterOptions.noVerboseOutputLimit && reqTextLen > bodyClipSize) {
```
shamasis, line 264: If the above recommendations are followed, then this if block become almost as original.

shamasis (changes requested):
We do not need to introduce two parameters for this. In the grand scheme of things this is a small feature and one parameter should do. Here's what we can do:

1. We ALWAYS clip. No matter what.
Providing Infinity as an input option can then automatically prevent clipping with least complicated codes. (And we mention this behaviour in docs as a short byline.)

2. We take only one option --reporter-cli-truncate-body-output (people will 99% chance refer documentation to get to this, hence we name accordingly.)

3. In the documentation, mention example input such as '1024', '1KB', etc so that users know what to input here and not need to refer the "bytes" readme doc.


PS - apologies for slow review, we are busy with a tonne of year change activities.

Sid200026:
@shamasis, I have simplified the implementation by using a single flag.

`newman run <collection> --verbose --reporter-cli-truncate-body-output <size in any unit>`. You can pass infinite to remove any truncation limit.

lib/reporters/cli/index.js
```diff
@@ -233,9 +235,24 @@ PostmanCLIReporter = function (emitter, reporterOptions, options) {
...
                 `${mime.charset}`
-            ].join(` ${colors.gray(symbols.star)} `);
-
+            ].join(` ${colors.gray(symbols.star)} `),
+            // eslint-disable-next-line max-len
+            bodyClipSize = BODY_CLIP_SIZE;
 
+        if (reporterOptions.truncateBodyOutput) {
+            // if infinite is passed to truncate-body-output
+            if (reporterOptions.truncateBodyOutput === 'infinite') {
```
shamasis, line 244: Why not take input as "infinity"? (Lowercase compare to ensure we don't need case sensitivity)

If you don't like the word infinity, then can use "none" too.

shamasis:
> @shamasis, I have simplified the implementation by using a single flag.
> 
> 
> 
> `newman run <collection> --verbose --reporter-cli-truncate-body-output <size in any unit>`. You can pass infinite to remove any truncation limit.

I think this is ready for approval. Feature now looks cool.

I've marked it as approved, but there's one tiny comment to resolve for you.

shamasis (dismissed):
Approved. One comment to be resolved.

lib/reporters/cli/index.js
Sid200026: I think infinity would be better since the description for --reporter-cli-truncate-body-output is specify the truncation size of body in verbose CLI output for requests and responses.

jmarcher:
I would love to have this

egbastos666:
Hello,

Is there any forecast to have this option published on newman? That's really something that will help a lot everyone, have the full request / response body written into a log. We work with images base64 inside the body and is impossible to check if everything was sent as expected

Eduardo

PR #2896, opened 2022-01-01
Added support for dynamic Iteration with Data Files

anshuljain03:
There are some changes pending on the current PR (#2652) which is open. This addresses those. 
Closes feature request raised in #2560

PS: This also has some extra whitespace removals

### DIFF
--- a/lib/run/index.js
+++ b/lib/run/index.js
@@ -142,6 +142,17 @@ module.exports = function (options, callback) {
             _.isArray(entrypoint.execute) && (entrypoint.lookupStrategy = MULTIENTRY_LOOKUP_STRATEGY);
         }
 
+        if (options.iterationData && (options.iterationDataFrom || options.iterationDataTo)) {
+            let dataSize = options.iterationData.length + 1,
+                start = options.iterationDataFrom,
+                end = options.iterationDataTo;
+
+            start = start >= 1 && start < dataSize ? start : 1;
+            end = end >= start && end <= dataSize ? end : dataSize;
+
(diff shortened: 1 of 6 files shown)


lib/run/index.js
```diff
@@ -142,6 +142,17 @@ module.exports = function (options, callback) {
...
 
+        if (options.iterationData && (options.iterationDataFrom || options.iterationDataTo)) {
+            let dataSize = options.iterationData.length + 1,
+                start = options.iterationDataFrom,
+                end = options.iterationDataTo;
+
+            start = start >= 1 && start < dataSize ? start : 1;
+            end = end >= start && end <= dataSize ? end : dataSize;
+
+            options.iterationData = options.iterationData.slice(start - 1, end);
```
shamasis, line 153: This needs to be done at `runtime` layer and not at Newman CLI layer for portability across usage in app and CLI. 

Additionally, care has to be taken that for ridiculously large data file size (not so uncommon) there isn't a memory usage spike.

shamasis (changes requested):
This is great effort. Really appreciate it.

This feature needs to be done at runtime layer and not at Newman CLI layer for portability across usage in app and CLI.

Additionally, care has to be taken that for ridiculously large data file size (not so uncommon) there isn't a memory usage spike.

shamasis:

All the above seems to be risky to implement even when it does what it is supposed to. These should be done at runtime level. Instead of manipulating data file, runtime itself should start and end on specific iterations. It would also help it make portable across CLI and App

PR #2908, opened 2022-01-20
Usage of `util.inherits()` is discouraged. 

k-rajat19:
`util.inherits()` is deprecated and should be avoided  to use. 
 see https://nodejs.org/api/util.html#utilinheritsconstructor-superconstructor


shamasis (approved):

k-rajat19:
Hi @shamasis , I think codecov status is stuck should i try adding new commits to trigger codecov report again ?
peoples are also facing similar issue -
https://community.codecov.com/t/codecov-status-stuck-at-waiting-for-status-to-be-reported-on-github/341/10

k-rajat19:
ping @codenirvana 🙂

k-rajat19:
@codenirvana  can u please  merge this

k-rajat19:
@codenirvana @shamasis  please merge this , i want to apply for Postman Contributor Program 😅

RaisinTen:
> `util.inherits()` is deprecated and should be avoided to use.

It's not deprecated, just legacy which means that it's okay if you use it. You can see the differences in https://nodejs.org/api/documentation.html#stability-index. The reason why it's marked legacy is that it doesn't establish the prototype chain as described in https://github.com/nodejs/node/issues/4179 but it works pretty well in most cases. Is this patch aiming to solve a problem we are facing with the usage of `util.inherits` here?

PR #2923, opened 2022-02-05
added support for proxy definitions

### LINKED ISSUE #1701: Feature request: support for proxy definitions
This is in line with a preceding use case here at https://github.com/postmanlabs/newman/issues/1698#issuecomment-418118274

One way would be to accept proxy list as part of CLI / file.

k-rajat19:
This PR can close #1701 which was opened as a feature request by @shamasis .
- [x] Feature
- [x] Update docs
- [ ] Add tests

Not sure about which proxy server  should be used for writing tests , should we need to create a proxy server using 
`http-proxy` or something else need to be done ?

Any suggestions/changes in this PR which leads to closing the above issue are welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -264,6 +264,13 @@ Newman can also be configured to work with proxy settings via the following envi
 
 For more details on using these variables, [refer here](https://github.com/postmanlabs/postman-request/blob/master/README.md#controlling-proxy-behaviour-using-environment-variables).
 
+#### Using Proxy configuration file
+
+- `--proxy-config-list`<br/>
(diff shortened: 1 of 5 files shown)


shamasis (changes requested):
This looks great. I have one suggestion.

We should use this opportunity to see how we can load these from a general config concept. Now, that may become slightly out of scope for this PR, hence to keep this sane and forward compatible, let's put the config's definition to be an array within an object.

```
{
  "proxies": [...]
}
```

PS: Meeting the test coverage (and changelog.yaml) will also be super cool and avoid someone else from needing to deep dive the PR post merge.

k-rajat19:
hi @shamasis , i have added new commits  also tried to adding some test cases as well
can you please look into this

k-rajat19:
@shamasis  @codenirvana
  is there any update  here :)

PR #2929, opened 2022-02-16
chore: replaced `substr()` with `substring()`

k-rajat19:
I think we should try to avoid using [deprecated](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) functions, just added a small fix for that in this PR.


k-rajat19:
@codenirvana  can u please review this PR if you have some time :)

PR #2934, opened 2022-02-22
Add docker image for Qase reporter using alpine image

AmmarArnt:
# What kind of change does this PR introduce?
Introduce docker file for building an image that uses the base alpine newman image, but with the Qase reporter installed. This enables the use of the newman container with Qase inside a kubernetes cluster, without the need to create separate node services or custom docker images in private repositories.

A similar approach could be used for other reporters.

# Does this PR introduce a breaking change?
No

# Other Information:
This image needs to be built and pushed to the dockerHub with the tag postman/newman:qase-alpine

### DIFF
--- a/docker/images/qase-alpine/Dockerfile
+++ b/docker/images/qase-alpine/Dockerfile
@@ -1,0 +1,3 @@
+FROM postman/newman:alpine
+
+RUN npm install --global newman-reporter-qase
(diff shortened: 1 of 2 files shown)


viktorlarsson:
Please merge this!

AmmarArnt:
Any feedback on this?

PR #2978, opened 2022-05-26
Feature: Added a new default reporter to publish newman run to postman

Harsh-Bansal:
This default reporter can be used to publish newman run to postman.
This reporter doesn't have any associated options.

### DIFF
--- a/lib/reporters/postman/helpers/constants.js
+++ b/lib/reporters/postman/helpers/constants.js
@@ -1,0 +1,70 @@
+/**
+ * An exhaustive set of constants used across various functions
+ */
+module.exports = {
+    /**
+     * Used as a source in the collection run object
+     */
+    NEWMAN_STRING: 'newman',
+
+    /**
+     * The status of the newman run in process
+     */
+    NEWMAN_RUN_STATUS_FINISHED: 'finished',
+
+    /**
+     * The success result of a particular test
+     */
+    NEWMAN_TEST_STATUS_PASS: 'pass',
+
+    /**
+     * The failure result of a particular test
+     */
+    NEWMAN_TEST_STATUS_FAIL: 'fail',
+
+    /**
+     * The skipped status of a particular test
+     */
+    NEWMAN_TEST_STATUS_SKIPPED: 'skipped',
+
+    /**
+     * Use this as a fallback collection name when creating collection run object
+     */
+    FALLBACK_COLLECTION_RUN_NAME: 'Collection Run',
+
+    /**
+     * The base URL for postman API
+     */
+    POSTMAN_API_BASE_URL: 'https://api.getpostman.com',
+
+    /**
+     * The API path used to upload newman run data
+     */
+    POSTMAN_API_UPLOAD_PATH: '/newman-runs',
+
+    /**
+     * Used as a fall back error message for the upload API call
+     */
+    RESPONSE_FALLBACK_ERROR_MESSAGE: 'Something went wrong while uploading newman run data to Postman',
+
+    /**
+     * Regex pattern to extract the collection id from the postman api collection url
+     */
+    COLLECTION_UID_FROM_URL_EXTRACTION_PATTERN: /https?:\/\/api\.getpostman.*\.com\/(?:collections)\/([A-Za-z0-9-]+)/,
+
+    /**
+     * Regex pattern to extract the environment id from the postman api environment url
+     */
+    ENVIRONMENT_UID_FROM_URL_EXTRACTION_PATTERN: /https?:\/\/api\.getpostman.*\.com\/(?:environments)\/([A-Za-z0-9-]+)/,
+
+    /**
+     * Regex pattern to extract the api key from the postman api collection url
+     */
+    API_KEY_FROM_URL_EXTRACTION_PATTERN:
+        /https:\/\/api.getpostman.com\/([a-z]+)s\/([a-z0-9-]+)\?apikey=([a-z0-9A-Z-]+)/,
+
+    /**
+     * Matches valid Postman UID, case insensitive.
+     */
+    UID_REGEX: /^[0-9A-Z]+-[0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{12}$/i
+};
--- a/lib/reporters/postman/helpers/run-utils.js
+++ b/lib/reporters/postman/helpers/run-utils.js
@@ -1,0 +1,248 @@
+const _ = require('lodash'),
+    uuid = require('uuid'),
+    {
+        NEWMAN_STRING,
+        FALLBACK_COLLECTION_RUN_NAME,
+        NEWMAN_RUN_STATUS_FINISHED,
+        NEWMAN_TEST_STATUS_PASS,
+        NEWMAN_TEST_STATUS_FAIL,
+        NEWMAN_TEST_STATUS_SKIPPED
+    } = require('./constants');
+
+/**
+ * Returns a request object that contains url, method, headers and body data
+ *
+ * Example reqeust object: {
+ *  url: 'https://postman-echo.com/get?user=abc&pass=123,
+ *  method: 'get',
+ *  headers: {
+ *    'Authorization': 'Basic as1ews',
+ *    'Accept': 'application/json'
+ *  },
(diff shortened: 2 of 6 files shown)


lib/run/index.js
```diff
@@ -42,7 +42,8 @@ var _ = require('lodash'),
         json: require('../reporters/json'),
         junit: require('../reporters/junit'),
         progress: require('../reporters/progress'),
-        emojitrain: require('../reporters/emojitrain')
+        emojitrain: require('../reporters/emojitrain'),
+        postman: require('../reporters/postman')
```
codenirvana, line 46: Are we sure the usage should be `-r postman`? What happened to `--publish`?

If yes, why it's named `postman`? 🤔

lib/reporters/postman/index.js
```diff
@@ -0,0 +1,76 @@
...
+ */
+function PostmanReporter (newman, _reporterOptions, collectionRunOptions) {
+    newman.on('beforeDone', (error, o) => {
+        if (error || !_.get(o, 'summary')) {
+            return;
+        }
+
+        // We get the collection id in the collectionRunOptions. But that collection id has the user id stripped off
+        // Hence, we try to parse the CLI args to extract the whole collection id from it.
+        const processArgs = process.argv,
```
codenirvana, line 22: How will this work when Newman is used as a library?

lib/reporters/postman/index.js
```diff
@@ -0,0 +1,76 @@
...
+            { collection, environment } = util.parseCLIArguments(processArgs),
+
+            // If api key is not present in environment variables, we check if it has been passed
+            // seperately as CLI args else we try to get it from collection postman api url
+            // eslint-disable-next-line no-process-env
+            postmanApiKey = process.env.POSTMAN_API_KEY ||
+                collectionRunOptions.postmanApiKey || util.getAPIKeyFromCLIArguments(processArgs);
+
+        if (!collection) {
+            print.lf('Publishing run details to postman cloud is currently supported only for collections specified ' +
```
codenirvana, line 32: Are we following the Newman logging pattern here?

lib/reporters/postman/helpers/constants.js
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Use this as a fallback collection name when creating collection run object
+     */
+    FALLBACK_COLLECTION_RUN_NAME: 'Collection Run',
+
+    /**
+     * The base URL for postman API
+     */
+    POSTMAN_API_BASE_URL: 'https://api.getpostman.com',
```
codenirvana, line 38: Why are we using the older `getpostman.com` domain name?

lib/reporters/postman/helpers/constants.js
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * The API path used to upload newman run data
+     */
+    POSTMAN_API_UPLOAD_PATH: '/newman-runs',
+
+    /**
+     * Used as a fall back error message for the upload API call
+     */
+    RESPONSE_FALLBACK_ERROR_MESSAGE: 'Something went wrong while uploading newman run data to Postman',
```
codenirvana, line 48: No, "Something went" error messages, please.

lib/reporters/postman/helpers/constants.js
```diff
@@ -0,0 +1,65 @@
...
+
+    /**
+     * Regex pattern to extract the environment id from the postman api environment url
+     */
+    ENVIRONMENT_UID_FROM_URL_EXTRACTION_PATTERN: /https?:\/\/api\.getpostman.*\.com\/(?:environments)\/([A-Za-z0-9-]+)/,
+
+    /**
+     * Regex pattern to extract the api key from the postman api collection url
+     */
+    API_KEY_FROM_URL_EXTRACTION_PATTERN:
```
codenirvana, line 63: Avoid this!

lib/reporters/postman/helpers/run-utils.js
```diff
@@ -0,0 +1,249 @@
...
+            return _.pick(header, ['key', 'value']);
+        });
+
+    return {
+        code: response.code,
+        name: response.status,
+        time: response.responseTime,
+        size: response.responseSize,
+        headers: headers,
+        body: response.stream ? new TextDecoder('utf-8').decode(new Uint8Array(response.stream)) : null
```
codenirvana, line 77: `response.text()`

codenirvana (changes requested):

lib/run/index.js
Harsh-Bansal: Using the -r postman follows the reporter convention which seems much better. Deprecating --publish would have been difficult (if needed).
Also, later when we add other configurable params (like timeout, retries), this approach would prevent bloating of newman cli options. They being specific to the reporter wont really belong as params of newman which doesnt use these (better separation of concerns on the interface)

Why `postman`: We went with postman as the data is being pushed to postman servers. (similar to other external reporters mentioned here (https://postmanlabs.atlassian.net/wiki/spaces/RUN/pages/3890544729))

We also considered `cloudpublish/postmanpublish/postmancloud` but `postman` seemed better.
Please let me know if you have any suggestions on this.

lib/reporters/postman/index.js
Harsh-Bansal: This will be available only via CLI.
The issue is we need the collection UID to publish runs (wrt a collection) and we can get it from CLI args but when used as a library we don't have a way to get it. (The reporter gets passed the collection object fetched via postman api which doesnt have the UID)
 
Will figure out a way to use it as part of library also in the second iteration.
(Our current target use-case is CLI invocations as part of CI/CD)

lib/reporters/postman/helpers/constants.js
Harsh-Bansal: @codenirvana we are using this only as a fallback in case apikey is not present as environment variable `POSTMAN_API_KEY` or passed as cli arg `--postman-api-key`

Lemmi know if that's fine.

lib/reporters/postman/helpers/constants.js
Harsh-Bansal: Switched to postman.com, thanks
Have also modified related regexs to support both `getpostman.com` and `postman.com`.

lib/reporters/postman/index.js
Harsh-Bansal: @codenirvana have made the changes to make the reporter compatible with library also.
cc @anshuljain03 @shubhbhargav

lib/util.js
```diff
@@ -53,7 +53,19 @@ var fs = require('fs'),
 
     // Matches valid Postman UID, case insensitive.
     // Same used for validation on the Postman API side.
-    UID_REGEX = /^[0-9A-Z]+-[0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{12}$/i;
+    UID_REGEX = /^[0-9A-Z]+-[0-9A-F]{8}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{4}-[0-9A-F]{12}$/i,
+
+    // Regex pattern to extract the collection id from the postman api collection url
+    COLLECTION_UID_FROM_URL_EXTRACTION_PATTERN =
```
shamasis, line 59: ReDos vulnerable RegEx

lib/util.js
```diff
@@ -53,7 +53,19 @@ var fs = require('fs'),
...
+        /https?:\/\/api\.(?:get)?postman.*\.com\/(?:collections)\/([A-Za-z0-9-]+)/,
+
+    // Regex pattern to extract the environment id from the postman api environment url
+    ENVIRONMENT_UID_FROM_URL_EXTRACTION_PATTERN =
+        /https?:\/\/api\.(?:get)?postman.*\.com\/(?:environments)\/([A-Za-z0-9-]+)/,
+
+    // Regex pattern to extract the api key from the postman api collection url
+    API_KEY_FROM_URL_EXTRACTION_PATTERN =
+        /https?:\/\/api.(?:get)?postman.com\/[a-z]+s\/[a-z0-9-]+\?apikey=([a-z0-9A-Z-]+)/;
```
shamasis, line 68: Seems not REDOS vulnerable 👍

shamasis (changes requested):

PR #3052, opened 2022-12-22
Fix cli reporter options table

smialkow:
There was a newline causing it to display incorrectly


smialkow:
dupe of https://github.com/postmanlabs/newman/pull/3011

Issue #3179, opened 2023-11-22
JSON loaded from external file doesn't unwrap when passed to programmatic runner


jfo:
When a collection is contained in an external JSON file, the behavior wrt how it is processed differs between the CLI interface and the programmatic interface in a surprising way.

When passed as a filepath, the collection can be loaded if it's in a JSON object OR if it's in a JSON object under the key `collection`, but when passed as JSON directly, this unwrapping doesn't happen. The unwrapping happens here:

https://github.com/postmanlabs/newman/blob/a612fbe29348eaae5f6aac0b882382d23d688ac3/lib/run/options.js#L60

This leads to surprising behavior, as my expectation was that running a json file that works on the CLI would also work when passed as part of a script. There is no feedback that the formatting of the json could be the issue and be causing the difference.

 1. Newman Version (can be found via `newman -v`): 6.0.0
 2. OS details (type, version, and architecture): Mac Ventura M2
 3. Are you using Newman as a library, or via the CLI? Both, see issue
 3. Did you encounter this recently, or has this bug always been there: Idk
 4. Expected behaviour: Consistency between CLI and programmatic use
 
 5. Command / script used to run Newman:
 6. Sample collection, and auxiliary files (minus the sensitive details):
reproduction 

Reproduction can be found in [this diff](https://github.com/postmanlabs/newman/compare/develop...jfo:newman:jfo/repro#diff-3848d513297c79fe6efd8a6ef0ba2dac7e2b3f3c87055204fb7da9369f9ede05) in [this branch](https://github.com/jfo/newman/tree/jfo/repro)

```
Wrapped JSON:  {
  iterations: { total: 0, pending: 0, failed: 0 },
  items: { total: 0, pending: 0, failed: 0 },
  scripts: { total: 0, pending: 0, failed: 0 },
  prerequests: { total: 0, pending: 0, failed: 0 },
  requests: { total: 0, pending: 0, failed: 0 },
  tests: { total: 0, pending: 0, failed: 0 },
  assertions: { total: 0, pending: 0, failed: 0 },
  testScripts: { total: 0, pending: 0, failed: 0 },
  prerequestScripts: { total: 0, pending: 0, failed: 0 }
}
Bare JSON:  {
  iterations: { total: 1, pending: 0, failed: 0 },
  items: { total: 3, pending: 0, failed: 0 },
  scripts: { total: 1, pending: 0, failed: 0 },
  prerequests: { total: 3, pending: 0, failed: 0 },
[...]

PR #3189, opened 2024-01-09
fix: JSON Collection Unwrapping Consistency in Programmatic Interface

Shiva953:
This pull request addresses an inconsistency in the programmatic interface of Newman related to the handling of JSON collections as described in #3179. When a collection is passed directly as JSON, there was unexpected behaviour compared to loading it from an external file, especially when unwrapping was required.

Changes Made:
- Modified the extractModel function in `newman/lib/run/options.js` to ensure consistent unwrapping behaviour for JSON collections.
- Added checks for the presence of the "values" property to distinguish between variables and other objects.


Shiva953:
would love a review for this, @codenirvana

PR #3224, opened 2024-04-25
update node to 20-alpine

mrbusche:


mrbusche:
need to update workflows

PR #3351, opened 2026-03-03
Chore - bump postman runtime to 7.52.0

JoaquimFlavio:


JoaquimFlavio:
It appears that the `timeout.script` option is not being respected in `postman-runtime@v7.52.0`, while `timeout.request` continues to function as expected.

I reviewed the documentation but couldn't find any mention of changes related to this behavior.

Could this be a regression in the current version? Should we wait for an upstream fix, or is there an alternative approach recommended?

> Please check if I'm not mistaken. 😅

PR #3368, opened 2026-08-02
Fix operator precedence bug dropping header size from total data received

### LINKED ISSUE #3367: Response header size dropped from "total data received" summary due to operator precedence bug
## Description

`RunSummary`'s request tracker computes `run.transfers.responseTotal` (shown in the CLI reporter as `total data received: ... (approx)`) using:

https://github.com/postmanlabs/newman/blob/develop/lib/run/summary.js#L315

```js
size && (summary.run.transfers.responseTotal += (size.body || 0 + size.headers || 0));
```

Because `+` binds tighter than `||` in JavaScript, this parses as:

```js
size.body || (0 + size.headers) || 0
```

not the intended:

```js
(size.body || 0) + (size.headers || 0)
```

## Impact

Whenever a response has a non-zero body size (i.e. almost always), `size.headers` is silently dropped from the running total — only `size.body` is counted. The "total data received" figure printed at the end of every collection run is therefore consistently undercounted by the size of all response headers across the run.

## Repro

```js
const size = { body: 270, headers: 793 };

console.log(size.body || 0 + size.headers || 0);        // 270  (headers silently dropped)
console.log((size.body || 0) + (size.headers || 0));    // 1063 (correct)
```

This is also confirmed by the existing test suite's own TODO in `test/unit/run-summary.test.js`:

```js
// @todo add test for computation of timings, transfer sizes and average response time
```

i.e. this computation currently has zero test coverage, which is how the bug went unnoticed.

## Fix

Add explicit parens so `+` happens before `||`:

```js
size && (summary.run.transfers.responseTotal += ((size.body || 0) + (size.headers || 0)));
```

PR: https://github.com/Adityaj0/newman/pull/2

Adityaj0:
## Summary

- `lib/run/summary.js`'s `attachRequestTracker` computed `run.transfers.responseTotal` with `size.body || 0 + size.headers || 0`. Since `+` binds tighter than `||` in JS, this evaluates as `size.body || (0 + size.headers) || 0`, so `size.headers` is silently dropped whenever `size.body` is truthy (i.e. almost always). The CLI's "total data received" summary line has therefore been undercounting by the size of every response's headers.
- Wrapped both terms in explicit parens: `(size.body || 0) + (size.headers || 0)`.
- Added regression tests covering single- and multi-request accumulation of `responseTotal`, closing the gap noted by the pre-existing `// @todo add test for computation of timings, transfer sizes and average response time` comment in `test/unit/run-summary.test.js`.

Fixes #3367

## Test plan

- [x] `node_modules/.bin/mocha test/unit/run-summary.test.js` — 70 passing, including 2 new tests
- [x] Confirmed the new tests fail against the old code (`270`/`300` instead of `1063`/`375`) and pass against the fix
- [x] `node_modules/.bin/eslint lib/run/summary.js test/unit/run-summary.test.js` — clean
- [x] `node_modules/.bin/mocha test/unit` — 130 passing, 11 pending (pre-existing, unrelated)

### DIFF
--- a/lib/run/summary.js
+++ b/lib/run/summary.js
@@ -312,7 +312,7 @@ _.assign(RunSummary, {
                 timingPhases;
 
             // compute the response size total
-            size && (summary.run.transfers.responseTotal += (size.body || 0 + size.headers || 0));
+            size && (summary.run.transfers.responseTotal += ((size.body || 0) + (size.headers || 0)));
 
             // if there are redirects, get timings for the last request sent
             timings = _.last(_.get(o, 'history.execution.data'));
--- a/test/unit/run-summary.test.js
+++ b/test/unit/run-summary.test.js
@@ -269,5 +269,54 @@ describe('run summary', function () {
                 });
             });
         });
+
+        describe('transfer size tracking', function () {
+            var emitter,
+                collection,
+                summary;
+
+            beforeEach(function () {
+                collection = new sdk.Collection({
+                    item: [{ id: 'i1', request: 'http://localhost/1' }]
+                });
+                emitter = new EventEmitter();
+                summary = new Summary(emitter, { collection });
+            });
+
+            afterEach(function () {
+                collection = null;
+                emitter = null;
+                summary = null;
+            });
+
+            it('should sum both body and header sizes into responseTotal', function () {
+                var item = collection.items.one('i1');
+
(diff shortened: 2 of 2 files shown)


Adityaj0:
@saecula can you review this too?

PR #3372, opened 2026-08-02
Fix JUnit reporter's testsuite tests attribute for multi-iteration runs

### LINKED ISSUE #3371: JUnit reporter's testsuite tests attribute is wrong (last iteration's count, not total unique assertions) on multi-iteration runs
## Description

The JUnit reporter groups every execution (i.e. every iteration) of the same collection item into a single `<testsuite>` element, and emits one `<testcase>` per unique assertion name seen across all of those iterations:

https://github.com/postmanlabs/newman/blob/develop/lib/reporters/junit/index.js#L47-L157

The `tests` attribute on that `<testsuite>`, however, is set *inside* the per-execution loop:

```js
if (execution.assertions) {
    suite.att('tests', execution.assertions.length);
}
else {
    suite.att('tests', 0);
}
```

Since this runs once per execution (i.e. once per iteration) and simply overwrites the attribute each time, the final value left on the suite after the loop is whatever the *last* iteration's assertion count happened to be — not the number of `<testcase>` elements actually generated for that suite.

## Impact

Whenever a collection is run with more than one iteration (`--iteration-count` / `-d` with multiple data rows) and the number of assertions differs between iterations (e.g. a conditional test script, or a test that only runs on the first iteration), the reported `tests` count is wrong and internally inconsistent with the emitted XML — CI systems that parse JUnit XML (Jenkins, GitLab, etc.) will see a `<testsuite tests="N">` that doesn't match its actual `<testcase>` child count, which can under/over-count executed tests in dashboards.

## Repro

```js
const EventEmitter = require('events');
const sdk = require('postman-collection');
const JunitReporter = require('./lib/reporters/junit');

const collection = new sdk.Collection({ item: [{ id: 'i1', name: 'Req1', request: 'http://localhost/1' }] });
const item = collection.items.one('i1');

const emitter = new EventEmitter();
emitter.exports = [];
emitter.summary = {
  collection,
  run: {
    executions: [
      { item, id: item.id, cursor: { iteration: 0 }, response: { responseTime: 10 }, assertions: [
          { assertion: 'status is 200', error: null },
          { assertion: 'body has id', error: null },
          { assertion: 'header is set', error: null }
      ] },
      { item, id: item.id, cursor: { iteration: 1 }, response: { responseTime: 10 }, assertions: [
          { assertion: 'status is 200', error: null }
      ] }
    ],
    timings: { started: Date.now() },
    stats: { tests: { total: 4 } }
  }
};

JunitReporter(emitter, {});
emitter.emit('beforeDone');
console.log(emitter.exports[0].content);
```

Output:

```xml
<testsuite name="Req1" id="i1" ... tests="1" failures="0" errors="0" ...>
  <testcase name="status is 200" .../>
  <testcase name="body has id" .../>
  <testcase name="header is set" .../>
</testsuite>
```

`tests="1"` but there are 3 `<testcase>` children.

## Fix

Move the `tests` attribute assignment out of the per-execution loop and set it once from the number of unique assertions actually collected:

```js
suite.att('tests', _.size(tests));
```

PR with fix + a unit test reproducing this (fails on current `develop`, passes with the fix): #3372

Adityaj0:
## Summary

- The JUnit reporter groups every iteration of a collection item into one `<testsuite>`, emitting one `<testcase>` per unique assertion name seen across all iterations.
- The `tests` attribute on that `<testsuite>` was set inside the per-execution loop to `execution.assertions.length`, so after the loop it ended up as whatever the *last* iteration's assertion count was, not the number of `<testcase>` elements actually emitted.
- On multi-iteration runs where the assertion count varies between iterations, this produces internally-inconsistent XML (`tests` doesn't match the actual `<testcase>` child count), which misreports test counts to any CI system parsing JUnit output.
- Moved the `suite.att('tests', ...)` call out of the loop, setting it once to `_.size(tests)` — the actual count of unique assertions collected, matching the `<testcase>` elements generated right below it.
- Added `test/unit/junit-reporter.test.js` reproducing the bug with two iterations that run a differing number of assertions.

Fixes #3371

## Test plan

- [x] `node_modules/.bin/mocha test/unit/junit-reporter.test.js` — passing
- [x] Confirmed the new test fails against unpatched `develop` (`tests: '1'` instead of `'3'`) and passes with the fix
- [x] `node_modules/.bin/eslint lib/reporters/junit/index.js test/unit/junit-reporter.test.js` — clean
- [x] `node_modules/.bin/mocha test/unit` — 131 passing, 11 pending (pre-existing, unrelated)
- [x] Existing single-iteration integration tests (`test/cli/shallow-junit-reporter.test.js`, `test/library/shallow-junit-reporter.test.js`) still pass unchanged — 11/11 passing

### DIFF
--- a/lib/reporters/junit/index.js
+++ b/lib/reporters/junit/index.js
@@ -110,16 +110,14 @@ JunitReporter = function (newman, reporterOptions) {
                         tests[name] = [];
                     }
                 });
-                if (execution.assertions) {
-                    suite.att('tests', execution.assertions.length);
-                }
-                else {
-                    suite.att('tests', 0);
-                }
 
                 suite.att('failures', failures);
                 suite.att('errors', errors);
             });
+
+            // one <testcase> is emitted per unique assertion name across all iterations of this item (see the
+            // forOwn loop below), so `tests` must reflect that count rather than a single iteration's tally
+            suite.att('tests', _.size(tests));
 
             suite.att('time', _.mean(_.map(executions, function (execution) {
                 executionTime = _.get(execution, 'response.responseTime') / 1000 || 0;
--- a/lib/run/summary.js
+++ b/lib/run/summary.js
@@ -312,7 +312,7 @@ _.assign(RunSummary, {
                 timingPhases;
 
             // compute the response size total
-            size && (summary.run.transfers.responseTotal += (size.body || 0 + size.headers || 0));
+            size && (summary.run.transfers.responseTotal += ((size.body || 0) + (size.headers || 0)));
 
             // if there are redirects, get timings for the last request sent
             timings = _.last(_.get(o, 'history.execution.data'));
--- a/test/unit/junit-reporter.test.js
+++ b/test/unit/junit-reporter.test.js
@@ -1,0 +1,80 @@
+const EventEmitter = require('events'),
+    expect = require('chai').expect,
+    xml2js = require('xml2js'),
+    sdk = require('postman-collection'),
+
+    JunitReporter = require('../../lib/reporters/junit');
+
+describe('junit reporter', function () {
+    var collection,
+        item,
+        emitter;
+
+    beforeEach(function () {
+        collection = new sdk.Collection({
+            item: [{ id: 'i1', name: 'Req1', request: 'http://localhost/1' }]
+        });
+        item = collection.items.one('i1');
+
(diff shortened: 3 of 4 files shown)


Adityaj0:
@appurva21 can you review this?

PR #3375, opened 2026-08-04
Newman v7

codenirvana:
**New Features:**
  - Added support for HTTP/2 via the new `--protocol-version` option
  - Uses the latest Postman Runtime with several improvements and bug fixes

**Breaking Changes:**
  - Dropped support for Node < v22.13
  - Dropped support for the v1 Postman Collection format, deprecated since v4
  - Requests now use HTTP/2 when the server supports it
  - Passing more than one collection to `newman run` is now an error instead of silently running the first
  - An unknown command now exits with code 1 instead of 0
  - An invalid option value now exits with code 1 instead of 0
  - Unquoted CSV iteration data values are now trimmed of all Unicode whitespace, such as non-breaking spaces

**Fixed Bugs:**
  - CSV iteration data with numeric-looking column names, such as `00123`, now loads instead of failing

**Chores:**
  - Upgraded Docker images to Node.js v24
  - Migrated to ESLint v10 flat config
  - Refreshed the ESLint rule set, formatting rules now come from `@stylistic` and Node.js rules from `eslint-plugin-n`
  - Updated dependencies


schwabyio:
Heads up: v7 pins postman-request@2.88.1-postman.49 → @postman/form-data@~3.1.1, which still calls the deprecated util.isArray (lib/form_data.js:64), so v7 will ship a DEP0044 warning on every multipart request. A one-line fix has been open at postmanlabs/form-data#5 since January, and since the range is ~3.1.1, merging it and publishing 3.1.2 would fix v7 automatically with no changes here.

